PR #18, opened 2023-09-11
Hover over anonymous struct or union fields results in odd __T<number> type name

### LINKED ISSUE #17: Hover over for anon structs and unions show internal type names
Give code like this:

```
typedef struct {
  int f;
} level_0;

typedef struct {
  level_0 f;
} level_1;

typedef struct {
  level_1 f;
} level_2;

void foo(level_2 x) {}
```

If you type a test script line in the editor like this:

```
TEST.VALUE:struct.foo.x.f.f
```
and hover over the "last f", you will see a typename like this: __T475666280 rather than level_0

johnpaliotta:
- Fix hover over for anonymous structs and unions show internal type names
- Fix stack trace display when VectorCAST environment version is incompatible with VectorCAST installation

Closes #17

### DIFF
--- a/python/tstUtilities.py
+++ b/python/tstUtilities.py
@@ -79,10 +79,11 @@ def getTypeDisplayName (type):
     """
 
     # The anon names are always __T<address> so use regex
-    if re.search("^__T[0-9]+$", type.display_name):
-        return type.typemark
-    else:
-        return type.display_name
+    returnValue = type.display_name
+    if re.match("__T[0-9]", returnValue):
+        returnValue = type.typemark
+    
+    return returnValue
 
 
 def additionalTypeInfo(type):
--- a/python/tstUtilities.py
+++ b/python/tstUtilities.py
@@ -66,21 +66,34 @@ def getObjectFromName(objectList, name):
     return None
 
 
+def getTypeDisplayName (type):
+    """
+    In some cases the "display_name" is a generated temporary or tag-name
+    this happens for anon structures in C for example
+    
+    typedef struct {
+        int a;
+        } myType;
+
+    In that case, use use the typemark vs display_name
+    """
+
+    # The anon names are always __T<address> so use regex
+    if re.search("^__T[0-9]+$", type.display_name):
+        return type.typemark
+    else:
+        return type.display_name
+
 
 def additionalTypeInfo(type):
     """
     For some types, we want to give a hint about the kind ...
     """
+
     if type.kind == "REC_ORD":
-        if re.search("^__T[0-9]+$", type.display_name):
-            # We have a generated temporary or tag-name
-            # this happens for anon structures in C for example
-            # typedef struct {int a;} myType;
-            return type.typemark + "(struct)"
-        else:
-            return type.display_name + "(struct)"
+        return getTypeDisplayName (type) + "(struct)"
     elif type.kind == "UNION":
-        return type.display_name + "(union)"
+        return getTypeDisplayName (type) +  "(union)"
     elif type.kind == "CLASS":
         return type.display_name + "(class)"
     elif type.kind in ["POINTER", "CLASS_PTR", "ACCE_SS"]:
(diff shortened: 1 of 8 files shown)


python/tstUtilities.py
```diff
@@ -66,15 +66,34 @@ def getObjectFromName(objectList, name):
...
+
 
 def additionalTypeInfo(type):
     """
     For some types, we want to give a hint about the kind ...
     """
+
     if type.kind == "REC_ORD":
-        return type.display_name + "(struct)"
+        return getTypeDisplayName (type) + "(struct)"
```
aytey, line 94: Are we aiming for f-strings in this code-base?

python/tstUtilities.py
```diff
@@ -66,15 +66,34 @@ def getObjectFromName(objectList, name):
...
+    
+    typedef struct {
+        int a;
+        } myType;
+
+    In that case, use use the typemark vs display_name
+    """
+
+    # The anon names are always __T<address> so use regex
+    if re.search("^__T[0-9]+$", type.display_name):
```
aytey, line 82: I'd just use `re.match` here without `^` and `$`.

Doing an `re.search` with a hard-coded `^` and `$` on a single-line string is the same as doing `re.match`.

python/vTestInterface.py
```diff
@@ -140,8 +140,13 @@ def generateTestInfo(test):
...
 def getTestDataVCAST(enviroPath):
 
-    api = UnitTestApi(enviroPath)
-
+    # dataAPI throws if there is a tool/enviro missmatch
+    try:
+        api = UnitTestApi(enviroPath)
+    except Exception as err:
+        print (err)
+        raise Exception("INVALID_ENVIRO")
```
aytey, line 148: Probably, the best way to do this is to make an `InvalidEnviro` exception and throw that.

At least, we should make `"INVALID_ENVIRO"` and `"USAGE_ERROR"` be string constants and re-use those rather than duplicating the strings.

python/tstUtilities.py
```diff
@@ -66,15 +66,34 @@ def getObjectFromName(objectList, name):
...
+        return type.typemark
+    else:
+        return type.display_name
```
aytey, line 85: You do do something like:

```python
ret = type.display_name
if re.match("^__T[0-9]+$", ret):
     ret = type.typemark
return ret
```

src/utilities.ts
```diff
@@ -359,7 +359,13 @@ export function executeCommand(
         .trim();
     else commandStatus.stdout = execSync(commandToRun).toString().trim();
   } catch (error: any) {
-    if (error && error.stdout) {
+    // 99 is a warning, like a missmatch opening the environment
```
aytey, line 362: mismatch (single s)

python/vTestInterface.py
```diff
@@ -140,8 +140,13 @@ def generateTestInfo(test):
 
 def getTestDataVCAST(enviroPath):
 
-    api = UnitTestApi(enviroPath)
-
+    # dataAPI throws if there is a tool/enviro missmatch
+    try:
+        api = UnitTestApi(enviroPath)
+    except Exception as err:
```
aytey, line 146: What does it actually throw?

python/tstUtilities.py
johnpaliotta: Done

python/vTestInterface.py
aytey: Done

python/tstUtilities.py
aytey: Not doing it

python/tstUtilities.py
aytey: Done

python/vTestInterface.py
johnpaliotta: File "vector/apps/DataAPI/migrations/migrate.py", line 44, in perform_necessary_migrations
```
raise MigrationError('''Master.db (version %i) is newer than your version of VectorCAST (%i).

aytey: Won't fix.

PR #19, opened 2023-09-14
Exception tidy-up

aytey:


aytey:
As a note: I've checked this is valid Python, but I haven't actually run it.

johnpaliotta (approved):

PR #23, opened 2023-10-04
Fix missing quotes in `--test` argument when forming `testInterfaceCommand`

Zbigor:
Running `testInterfaceCommand` in E2E tests failed with `Unterminated quoted string`
- After adding the missing quote, all tests pass
- Committed changes to expected output of E2E tests
  - Relying on terminal output here, hence `vcast:` in the log

### DIFF
--- a/tests/internal/e2e/test/specs/vcast.test.ts
+++ b/tests/internal/e2e/test/specs/vcast.test.ts
@@ -491,7 +491,7 @@ describe("vTypeCheck VS Code Extension", () => {
     ).toBe(true);
     expect(
       outputViewText.includes(
-        "test explorer  [info]  Test summary for: cpp/unitTests/DATABASE-MANAGER|manager.Manager::PlaceOrder.myFirstTest",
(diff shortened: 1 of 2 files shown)


tests/internal/e2e/test/specs/vcast.test.ts
```diff
@@ -491,7 +491,7 @@ describe("vTypeCheck VS Code Extension", () => {
     ).toBe(true);
     expect(
       outputViewText.includes(
-        "test explorer  [info]  Test summary for: cpp/unitTests/DATABASE-MANAGER|manager.Manager::PlaceOrder.myFirstTest",
+        "test explorer  [info]  Test summary for: vcast:cpp/unitTests/DATABASE-MANAGER|manager.Manager::PlaceOrder.myFirstTest",
```
aytey, line 494: @johnpaliotta didn't we say that the `vcast:` prefix was something "internal and hidden" and wasn't supposed to be user-visible?

johnpaliotta: It is not "seen" except in this logging context, which I think is ok

PR #24, opened 2023-10-06
Avoid overwriting CCAST_.CFG

### LINKED ISSUE #25: Extension overwrites existing `CCAST_.CFG`
If you already have a `CCAST_.CFG` in your `unitTests` folder and then you go to create a new VectorCAST environment, the extension will overwrite your existing `CCAST_.CFG` with the template `GNU_CPP_X`.

This is a bigger issue if your `CCAST_.CFG` contains search paths (etc.) because then you lose these, meaning creating a new environment can fail due to missing search paths.

Reported by @Waseem-Ashraf

aytey:
This PR changes `vector-vscode-vcast` such that we do not overwrite an existing `CCAST_.CFG` if one already exists.

Before this PR:

* `buildEnvironmentVCAST` -> always creates a `CCAST_.CFG` (overwriting an existing one) -> calls `executeClicastCommand` -> `executeClicastCommand` checks for a `CCAST_.CFG` and creates one if it doesn't exist (but it does, because `buildEnvironmentVCAST` always makes one)

After this PR:

* `buildEnvironmentVCAST` -> calls `executeClicastCommand` -> `executeClicastCommand` checks for a `CCAST_.CFG` and creates one if it doesn't exist (makes one if it doesn't exist, doesn't touch it if does)

Closes #25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,3 +60,9 @@ This option is new for VectorCAST 23 sp2
 - Fixed issue:  New test scripts should have TEST.NEW not TEST.REPLACE #20
 - Fixed issue:  Added ENVIRO.STUB: ALL_BY_PROTOTYPE to environment script #21
 - Fixed issue:  Execute error when enviro at root of workspace #22
+
+## [1.0.7] - 2023-10-06
+
+### Bug Fixes
+- Fixed issue: Do not overwrite an existing `CCAST_.CFG` if one exists #25
+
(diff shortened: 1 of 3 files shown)


johnpaliotta (approved):

johnpaliotta:
> This PR changes `vector-vscode-vcast` such that we do not overwrite an existing `CCAST_.CFG` if one already exists.


@aytey Thanks for fixing this, can you also create an issue for this.

Issue #30, opened 2023-10-11
[New Feature] Allow right click VectorCAST actions for .vce and .env files


johnpaliotta:
- .vce right click should open the environment
- env right click should build a new environment

PR #33, opened 2023-10-13
New context menu choices for .env and .vce files

johnpaliotta:
- Implemented Issue #30


Zbigor (approved):
Looks good now, passed e2e tests

PR #35, opened 2023-10-17
Add RGW support to test script editor

johnpaliotta:
Complete support has been added for TEST.REQUIREMENT_KEY licenese, including:
- Syntax highlighting
- Auto-complete shows all available requirements
- Hover over shows title and description
- E2E tests

### DIFF
--- a/syntax/vcast.tst.grammar.json
+++ b/syntax/vcast.tst.grammar.json
@@ -15,7 +15,7 @@
       "name": "keyword"
     },
     {
-      "match": "^TEST\\.(?:UNIT|SUBPROGRAM|NAME|SCRIPT_FEATURE|STUB)(:|$)(.*)$",
+      "match": "^TEST\\.(?:UNIT|SUBPROGRAM|NAME|SCRIPT_FEATURE||REQUIREMENT_KEY)(:|$)(.*)$",
       "name": "keyword",
       "captures": {
         "1": {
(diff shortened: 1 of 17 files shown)


johnpaliotta:
@aytey I updated the launch.json because the syntax of the command changed.

syntax/vcast.tst.grammar.json
```diff
@@ -15,7 +15,7 @@
       "name": "keyword"
     },
     {
-      "match": "^TEST\\.(?:UNIT|SUBPROGRAM|NAME|SCRIPT_FEATURE|STUB)(:|$)(.*)$",
+      "match": "^TEST\\.(?:UNIT|SUBPROGRAM|NAME|SCRIPT_FEATURE||REQUIREMENT_KEY)(:|$)(.*)$",
```
aytey, line 18: Accidental `||` in here?

aytey: Also, where did STUB go?

johnpaliotta: fixed

johnpaliotta:
@Zbigor  @aytey Tests look good.  Should we / can we test the work I do on right click load?  As you know when there is a requirement that was entered via the extension, it gets put in like this:
```
TEST.REQUIREMENT_KEY:FR20 | Clearing a table resets orders for all seats
```
On the load action I turn that into a comment containing the title part and a second line: TEST.REQUIREMENT_KEY:FR20
So you could test this two ways, ensure that the TEST.REQUIREMENT_KEY:FR20 line is there, or ensure that clicast can load the script without any errors.

Issue #31, opened 2023-10-12
[New feature] Add a context menu "New Basis Path Test Script"


joyyuansh:
In GUI, "Insert Basis  Path Test Cases" are generally the start point for most of the users. Is it practical to add a context menu for test environment/UUT/subprogram in the Test Explorer to automatically generate a test script file which contains all the basis path test cases? So that the users can generate the basis path test cases in Test Explorer, otherwise, users have to do three steps: 1. open the environment; 2. insert basis path test cases 3. close VectorCAST GUI

PR #38, opened 2023-10-20
Re-organized code to group most of the clicast commands into a new file

johnpaliotta:
Reorganization of the the clicast logic into a separate file


johnpaliotta:
@aytey @Zbigor should be no _real_ changes wanted to do this before adding more clicast commands (issue #31 for example)

src/clicast.ts
```diff
@@ -0,0 +1,264 @@
...
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
+const spawn = require("child_process").spawn;
+
+
+const clicastName = "clicast";
+export let clicastCommandToUse: string | undefined = undefined;
+
+const vcastqtName = "vcastqt";
```
aytey, line 35: Can we call this package something else, given it isn't _just_ `clicast`?

src/clicast.ts
```diff
@@ -0,0 +1,264 @@
...
+    else {
+      newLines.push(line);
+    }
+  }
+  fs.writeFileSync(scriptPath, newLines.join("\n"), "utf8");
+}
+
+
+const url = require("url");
+export async function loadTestScript() {
```
Zbigor, line 149: We have `loadTestScript()`, `deleteTest()` and others that represent testPane interactions. We also have "low-level" `clicast` - related functions like `executeClicastCommand()`. Maybe we could leave the "low-level stuff" in this file and move the "test pane interactions" to a separate file (maybe there are some other interaction functions still in `testPane.ts` at the moment?)

src/utilities.ts
```diff
@@ -30,15 +29,6 @@ export let globalTestInterfacePath: string | undefined = undefined;
 let globalCrc32Path: string | undefined = undefined;
 let globalPathToSupportFiles: string | undefined = undefined;
 
-export function setPaths(
```
Zbigor, line 33: We use `setPaths` for unit tests, but from `pythonUtilities`. I guess it's ok to remove it from here

src/clicast.ts
johnpaliotta: yea, I considered that, but the only vcastqt thing is the name, will re-consider

src/utilities.ts
johnpaliotta: Ohh, wondered why this was not called :)  If you need this, we syould add it back into clicast.ts, want me to do that?

Zbigor: I checked now. Tests use `setPaths` from `server/pythonUtilities.ts`. No need to add this back to `clicast.ts`

src/clicast.ts
johnpaliotta: Will change to vcastUtilities.py

johnpaliotta: Done

src/clicast.ts
johnpaliotta: Ok I will use the convention that the commands are in testPane, and the clicast related logic is in vcastUtilities.

Let me know what you think

Zbigor (approved):
Up-to-date with main now and e2e tests passed

PR #39, opened 2023-10-20
E2E tests for new context menu choices

Zbigor:
- Added a test for building an environment from a `.env` file
- Added a test for starting `vcastqt` from the `.vce` file
- Improved robustness of other e2e tests


johnpaliotta (approved):

johnpaliotta:
Don't know everting about these tests, but this looks right!

PR #40, opened 2023-10-20
Allow automatic test generation

johnpaliotta:
Add a new right click menu item to the VectorCAST context menu in the test explorer, to allow the generation of automatic tests, such as basis path tests.


aytey:
@Zbigor should we bump to 1.78 in this PR?

Zbigor:
> @Zbigor should we bump to 1.78 in this PR?

Yeah, we could do that here since the functionalities introduced in this PR require 1.78, makes sense

aytey:
@johnpaliotta does this make sense:

image

That is, if you click on a "test" node, you get an option to insert ATG or basis path tests _for the test_? I would have expected this menu to only exist at the unit or subprogram level.

aytey:
I think this is "broken" for the case where you have parameterisation.

From within VS Code:

image

On the VS Code debug output:

image

Here's now it should look:

(quoted name):

image

(escaped name):

image

If you don't quote or escape:

image

aytey:
@Zbigor can you make a child branch of this branch, and add:

```cpp
class Moo {
private:
  void honk(int x) {}

public:
  void honk(int x, int y) {}
  void honk(int x, int y, int z) {}
};
```

as a test example for ATG (right click on `Moo::honk(int,int,int)` and try to add tests)? Today, I think this is going to fall over (which is good, because then we have the test when the bug is fixed).

johnpaliotta:
> @johnpaliotta does this make sense:
> 
> image
> 
> That is, if you click on a "test" node, you get an option to insert ATG or basis path tests _for the test_? I would have expected this menu to only exist at the unit or subprogram level.

Ya it is really hard to turn context menus on/off because this is controlled by checks that are in the package.json, and the checks can only be made against special variables that are "pushed" from the typescript.  I currently push the enviro node names to a flat list, but it is not simple to push the "test names" since names can be shared etc.  We might be able to figure out a complicated solution but not in this branch.

johnpaliotta:
> I think this is "broken" for the case where you have parameterisation.
> 
> From within VS Code:
> 
> image
> 
> On the VS Code debug output:
> 
> image
> 
> Here's now it should look:
> 
> (quoted name):
> 
> image
> 
> (escaped name):
> 
> image
> 
> If you don't quote or escape:
> 
> image

I think u fixed this right?

aytey:
> I think u fixed this right?

Yes, I believe it should be fixed in https://github.com/vectorgrp/vector-vscode-vcast/pull/43.

I'd like @Zbigor to write the test so we can confirm that.

PR #42, opened 2023-11-10
E2E tests for automatic test generation features

Zbigor:
For both ATG and BASIS PATH added tests for:
- right-click on environment -> insert tests
- right-click on unit -> insert tests
- right-click on function -> insert tests
- right-click on flask+ icon -> generate tests
- multi-select delete
- deletion at environment, unit and function level

The ATG tests can be enabled by setting `process.env["ENABLE_ATG_FEATURE"]` 

**Note - we need to bump the required vs-code version to 1.78.0** (in another PR)
- Menu for inserting tests does not appear until vs-code version 1.77.0 (manually reproduced)
- Menu for generating tests after clicking on flask+ icon does not appear until vs-code version 1.78.0 (manually reproduced)


johnpaliotta:
@Zbigor as discussed with Andrew, please change the enviro variable to be something like

process.env["ENABLE_ATG_FEATURE"] = "TRUE"

Zbigor:
> process.env["ENABLE_ATG_FEATURE"]

Done

aytey (approved):

Issue #58, opened 2024-01-31
"Create VectorCAST Environment" command is not working in some cases.


johnpaliotta:
Issue happens in the case where the extension cannot find a VectorCAST installation.
- VectorCAST is not on path
- $VECTORCAST_DIR is not set
- Workspace setting: "VectorCAST Installation Location" is not set

In these cases, if you open a workspace that contains a VectorCAST environment, or if you activate the command: "VectorCAST Test Explorer: Configure" from the Command Palette, or if you right click on a C/++ source file and choose "Create VectorCAST Environment" the extension will look for a VectorCAST installation, and not finding one will display a series of status messages indicating that the VectorCAST installation was not found and open the Settings pane to allow you to configure the Installation Location.

There are two bugs:
- When the Installation Location option is changed, it is not validated
- The Create VectorCAST Environment command does nothing

Workaround: after updating the Installation Location, use the "VectorCAST Test Explorer: Configure" to validate the installation

johnpaliotta:
Test cases
- VectorCAST not on path, $VECTORCAST_DIR not set
  - Open a folder with no VectorCAST environments, and VectorCAST Installation Location not set
    - Right click on foo.cpp -> Create VectorCAST Environment -> Output pane shows error, settings opens
    - Set VectorCAST Install Location to an invalid path -> Output pane shows error
    - Right click on foo.cpp -> Create VectorCAST Environment -> Output pane shows error
    - Set VectorCAST Install Location to a valid path -> Output pane shows good status
    - Right click on foo.cpp -> Create VectorCAST Environment -> Environment Builds -> Test Pane Shows FOO environment
    - Delete the .vscode/settings.json folder
    - Close Folder
  - Open same folder -> Output pane shows error, settings opens
    - Set VectorCAST Install Location to a valid path -> Output pane shows good status
    - Set VectorCAST Install Location to a valid path -> Output pane shows good status
      - Flask appears with FOO environment shown

PR #59, opened 2024-01-31
Fix for issue #58, "Create VectorCAST Environment" not working in some cases

johnpaliotta:


src/extension.ts
```diff
@@ -551,37 +565,79 @@ function configureExtension(context: vscode.ExtensionContext) {
...
+
+      // contains a check for already configured, so no work will be done in that case
+      checkPrerequisites(context)
+      if (alreadyConfigured) {
+        // arg is the actual item that the right click happened on, argList is the list
+        // of all items if this is a multi-select.  Since argList is always valid, even for a single
+        // selection, we just use this here.
+        if (argList) {
+          newEnvironment(argList);
+        }
```
Zbigor, line 632: @johnpaliotta Are we sure we kept the desired behavior when `alreadyConfigured` is still false when making the new environment? It seems correct to me, since we only had `if (argList) {newEnvironment(argList);}` in 
`"vectorcastTestExplorer.newEnviroVCAST"` before and things were being triggered properly, but I think it'd make sense to double check

Zbigor (commented):
Overall logic looks good (would be good to just double check the part I commented on). The tests are passing

src/extension.ts
johnpaliotta: Yes, this is right because check prerequisites will either "pass" and set the already configured, or will fail and open the settings and output pane so you can fix things.

PR #55, opened 2024-01-22
Coded Test Support

johnpaliotta:
Added support for the new Coded Test feature that will be released with VectorCAST 24.  These changes have no affect on users of older VectorCAST versions, or users who do not wish to use Coded Tests.  The usage of the Coded Test features is described in the README.md.   The following is a summary of the new features:

- Building an environment with Coded Test support.
- Generating a new Coded Test source file for any unit
- Adding an existing Coded Test source file for any unit
- Editing Coded Test source files, including IntelliSense support
- Update of the test explorer tree to display new tests as they are added in the editor
- Execution of Coded Tests from the test explorer tree, or by using the gutter icons in the source editor
- Support for debugging Coded Tests

PR #63, opened 2024-02-06
Fix for issue: #64, improve colors for user code expected values

aytey:
When testing #55, I noticed that the failure colours for coded tests looked very bad in dark mode. The issue was that the error colour was "too light", while the text was also light, making it very hard to read.

This PR changes the extension such that we correctly handle `.bg-danger`, as well as choosing a slightly more "punchy" colour for the error colour (which I feel works better between the two themes).

### DIFF
--- a/src/reporting.ts
+++ b/src/reporting.ts
@@ -53,8 +53,8 @@ tr:hover {
     overflow: scroll;
 }
 
-.danger {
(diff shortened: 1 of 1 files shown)


aytey:
## Before

### Light

before_light

### Dark

before_dark

## After

### Light

after_light

### Dark

after_dark

Issue #67, opened 2024-02-13
Right click build environment using an .env file can cause environment deletion


johnpaliotta:
If you have an environment FOO, and the associated FOO.env, right clicking on FOO.env and choosing: Build VectorCAST Environment results in an error because the environment already exists, but worse, it deletes the existing FOO environment directory.

PR #68, opened 2024-02-13
Fix for issue #67, right click on .env file and choosing build

johnpaliotta:
VectorCAST environment, deletes an existing environment
@Zbigor can you add a quick test for this.


Zbigor (approved):
All good

PR #70, opened 2024-02-14
Improve atg messages

johnpaliotta:
The messages emitted when adding ATG tests have extra \n characters, like this:
test explorer  [info]  Subprogram: Manager::PlaceOrder(unsigned,unsigned,OrderType)  
1.
.
.
.

This has been improved the strip the extra \n, so the messages look like this:
test explorer  [info]  Subprogram: Manager::PlaceOrder(unsigned,unsigned,OrderType)  1.................. 2.................. 3.........


Zbigor (approved):
image

I assume this is what we expect, like a "progress bar per generated test". Looks fine to me. Had a look at the code, logic looks ok as well

PR #73, opened 2024-02-29
Move to 1.0.11 for next release

aytey:


aytey:
This is to avoid us giving "fixes" to customers and the fixes still looking like 1.0.10 -- so, the first thing to do after a new release, is bump the version tag!

Zbigor (approved):

PR #74, opened 2024-03-04
Improve JSON data handling

johnpaliotta:
Removed getCoverageData command from the vTestInterface python API since the coverage data is already returned by the getEnviroData command. Also did some other code cleanup for consistency


Zbigor (approved):
Test ran fine, change looks good

Issue #76, opened 2024-03-05
The Unit Test Location option does not support relative paths with nesting


johnpaliotta:
Values like "./tests/a/b/c" are not supported, but values like: "./tests" and "/home/user/tests/a/b/c" are supported.

When the value is set to "./tests/a/b/c" a build environment action will prompt to create the directory, but will do nothing and provide no warning message.

aytey:
> and "/home/user/tests/a/b/c" are supported.

What happens if `/home/user/tests` doesn't exist when trying to use `/home/user/tests/a/b/c`?

johnpaliotta:
> > and "/home/user/tests/a/b/c" are supported.
> 
> What happens if `/home/user/tests` doesn't exist when trying to use `/home/user/tests/a/b/c`?

We catch it when you set the option value

PR #77, opened 2024-03-05
Fix for issue #76, nested relative paths for unit test location

johnpaliotta:
There are two cases handled in the fix
- An illegal value for the path, like: "./unitTests/****"
- A valid value with nested syntax like: "./unittests/a/b/c

@Zbigor both should be tested

### DIFF
--- a/src/vcastTestInterface.ts
+++ b/src/vcastTestInterface.ts
@@ -656,9 +656,9 @@ function configureWorkspaceAndBuildEnviro(
           try {
             fs.mkdirSync(unitTestLocation, { recursive: true });
             commonNewEnvironmentStuff(fileList, unitTestLocation);
-          } catch (error) {
+          } catch (error:any) {
             vscode.window.showErrorMessage(
-              `Error creating directory: ${unitTestLocation}, update the 'Unit Test Location' option to a valid value`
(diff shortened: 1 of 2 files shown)


johnpaliotta:
@aytey Igor created the test and is passes, any comments on the changes here, or can we merge?

src/vcastTestInterface.ts
```diff
@@ -653,8 +653,16 @@ function configureWorkspaceAndBuildEnviro(
...
       .then((answer) => {
         if (answer === "Yes") {
-          fs.mkdirSync(unitTestLocation);
-          commonNewEnvironmentStuff(fileList, unitTestLocation);
+          try {
+            fs.mkdirSync(unitTestLocation, { recursive: true });
+            commonNewEnvironmentStuff(fileList, unitTestLocation);
+          } catch (error) {
+            vscode.window.showErrorMessage(
+              `Error creating directory: ${unitTestLocation}, update the 'Unit Test Location' option to a valid value`
```
aytey, line 661: Is there a chance there's anything in `error` that we should be showing to the user? For example, if it is a permissions problem?

johnpaliotta: Good point, I had not tested a permissions error, dialog looks like this now

image

johnpaliotta: Nope

PR #79, opened 2024-03-08
e2e test for relative paths with nesting

Zbigor:


Zbigor:
> @Zbigor Can you verify that the directory ./unitTests/a/b/c gets created and that the .env file is located there?

Done

johnpaliotta (approved):

PR #82, opened 2024-03-18
Fix colour scheme for stdout/stderr from driver (`<pre>`)

aytey:
This PR fixes the CSS styling for `<pre>` blocks, which are used in reports for stdout/stderr from VectorCAST's harness.


aytey:
Before:

image

After:

image

johnpaliotta (approved):

Issue #81, opened 2024-03-15
Existing environment directories are not found if locked directories exist in the workspace


johnpaliotta:
If there are directories in the workspace that do not have read permission then, the harvesting of existing VectorCAST environment data will fail, and the test tree will be empty.

PR #83, opened 2024-03-19
Fix for issue: #81 and some status improvements

johnpaliotta:
Fixed issue: #81 and added some new status messages when we are doing long running coded-test actions.

### DIFF
--- a/src/testPane.ts
+++ b/src/testPane.ts
@@ -300,7 +300,7 @@ function getEnvironmentList(baseDirectory: string): string[] {
   // This function will find all of the VectorCAST and vTest
   // environments downstream of the current workspace
 
-  const options = { cwd: baseDirectory, absolute: true };
+  const options = { cwd: baseDirectory, absolute: true, strict: false };
   let fileList = glob.sync("**/" + vcastEnviroFile, options);
 
   // now we have a list of the UNITDATA.VCD files downstream of us
@@ -661,7 +661,7 @@ function getFileToDebug (
     }
   }
 
-  const globOptions = { cwd: enviroPath, absolute: true };
+  const globOptions = { cwd: enviroPath, absolute: true, strict: false };
   // two steps for debugging ...
   const globResult = glob.sync(globPattern, globOptions);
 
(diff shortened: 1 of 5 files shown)


src/testPane.ts
```diff
@@ -1273,8 +1273,8 @@ export function updateCodedTestCases (editor:any) {
         // note: the same file should never be associated with more than one unit
         let commandToRun: string = 
           `${clicastCommandToUse} ${getClicastArgsFromTestNode(testNode)} test coded refresh`;
+        vectorMessage (`Refreshing coded test file: ${filePath} for environment: ${enviroPath}`);
```
Zbigor, line 1276: This PR changes `vectorMesage` to be async. We have many places in the project where we call `vectorMessage` inside a **non-async** function. This means that the function execution will continue right away and will not wait for message display to be done. We cannot "await" the `vectorMessage` inside a non-async functions, so the message can not be displayed before a long running process. 

Is this a desired behavior? If we need an async variant of `vectorMessage` for use in particular **async functions**, we could implement it separately, while still keeping the synchronous variant we had before

src/testPane.ts
```diff
@@ -661,7 +661,7 @@ function getFileToDebug (
     }
   }
 
-  const globOptions = { cwd: enviroPath, absolute: true };
+  const globOptions = { cwd: enviroPath, absolute: true, strict: false };
```
Zbigor, line 664: I guess setting strict to false is what solves the issue?

src/testPane.ts
johnpaliotta: I'm pretty sure that the single function will operate this way ... if called with await (from an aysnc function) the message will be forced to be displayed and otherwise it will work as it did before.  This gives us the option to make the caller async in cases where we want the message displayed, and that's what I did for the two lambda functions.

src/testPane.ts
johnpaliotta: yep

src/testPane.ts
Zbigor: ok, all good then

Zbigor (approved):

Issue #78, opened 2024-03-05
Run with debug overwrites JSON-C based launch.json files.


johnpaliotta:
If JSON-C (JSON with comments) is used for the launch.json file, the extension will over-write the contents on a "Debug Test" action.

Workaround is the remove comments and trailing commas from the launch.json

PR #84, opened 2024-03-21
Added support for JSON-C formatted configuration files

johnpaliotta:
This PR fixes issue: #78

Updated the processing of launch.json, c_cpp_properties.json, and settings.json to use jsonc-parser which supports reading and writing json-c files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -569,7 +569,7 @@
   },
   "scripts": {
     "vscode:prepublish": "npm run esbuild-base -- --minify && npm run esbuild-server -- --minify",
-    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/extension.js --external:vscode --format=cjs --platform=node",
+    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/extension.js --external:vscode --external:jsonc-parser --format=cjs --platform=node",
     "esbuild-server": "esbuild ./server/server.ts --bundle --outfile=out/server.js --external:vscode --format=cjs --platform=node",
     "esbuild-watch": "npm-run-all --parallel esbuild-base esbuild-server -- --sourcemap --watch",
     "esbuild": "npm run esbuild-base -- --sourcemap && npm run esbuild-server -- --sourcemap",
@@ -606,7 +606,8 @@
     "npm-run-all": "^4.1.5",
     "tslib": "^1.9.3",
     "vscode-languageclient": "^5.2.1",
-    "vscode-languageserver": "^5.2.1"
+    "vscode-languageserver": "^5.2.1",
+    "jsonc-parser": "^3.2.1"
   },
   "extensionDependencies": [],
   "engines": {
--- a/src/testPane.ts
+++ b/src/testPane.ts
@@ -571,9 +571,9 @@ function launchConfigExists(launchJsonPath: string): boolean {
   // this will check if launch.json has
   // the "VectorCAST Harness Debug" configuration defined
   let returnValue = false;
-  const existingJSON: any = loadLaunchFile(launchJsonPath);
-  if (existingJSON.configurations) {
-    for (const existingConfig of existingJSON.configurations) {
+  const existingJSONdata: any = loadLaunchFile(launchJsonPath);
+  if (existingJSONdata) {
+    for (const existingConfig of existingJSONdata.jsonData.configurations) {
       if (existingConfig.name === vectorcastLaunchConfigName) {
         returnValue = true;
         launchFilesWithVectorCASTConfig.set(launchJsonPath, true);
@@ -719,11 +719,16 @@ async function debugNode(
 
         createEmptyLaunchConfigFile(ourWorkspace, launchJsonPath);
         addLaunchConfiguration(launchJsonUri);
-      } else {
+      } 
+      else {
         debugConfigurationFound = launchConfigExists(launchJsonPath);
-        if (!debugConfigurationFound) addLaunchConfiguration(launchJsonUri);
+        if (!debugConfigurationFound) {
+          addLaunchConfiguration(launchJsonUri);
+        }
       }
 
+      // this flag means that the launch file already had a vectorcast debug configuration
+      // if we just added it we want to give the user a chance to review before we debug
       if (debugConfigurationFound) {
         vectorMessage(
           `Preparing to debug test ${getTestNameFromID(node.id)} ... `
--- a/src/utilities.ts
+++ b/src/utilities.ts
@@ -1,7 +1,7 @@
 import * as vscode from "vscode";
 
 // TBD TODAY - import
-import * as jsonc from "jsonc-parser"
+//import * as jsonc from "jsonc-parser"
 
 import { Uri } from "vscode";
 
@@ -19,7 +19,8 @@ const vPythonName = "vpython";
 const vpythonFromPath = which.sync(vPythonName, { nothrow: true })
 export let vPythonCommandToUse: string | undefined = undefined;
 
-export const jsoncParseOptions:jsonc.ParseOptions = { allowTrailingComma: true, disallowComments: false, allowEmptyContent:false };
+// TBD TODAY - parseOptions
+//export const jsoncParseOptions:jsonc.ParseOptions = { allowTrailingComma: true, disallowComments: false, allowEmptyContent:false };
 
 
 // The testInterface is delivered in the .vsix
--- a/src/utilities.ts
+++ b/src/utilities.ts
@@ -19,9 +19,8 @@ const vPythonName = "vpython";
 const vpythonFromPath = which.sync(vPythonName, { nothrow: true })
 export let vPythonCommandToUse: string | undefined = undefined;
 
-// TBD TODAY - parseOptions
+// options used for reading json-c files
 export const jsoncParseOptions:jsonc.ParseOptions = { allowTrailingComma: true, disallowComments: false, allowEmptyContent:false };
-
 
 // The testInterface is delivered in the .vsix
 // in the sub-directory "python"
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -569,7 +569,7 @@
   },
   "scripts": {
     "vscode:prepublish": "npm run esbuild-base -- --minify && npm run esbuild-server -- --minify",
-    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/extension.js --external:vscode --format=cjs --platform=node",
+    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/extension.js --external:vscode --external:jsonc-parser --format=cjs --platform=node",
```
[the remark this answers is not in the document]
johnpaliotta: It gets installed into the node_modules the external allows the path stuff to resolved properly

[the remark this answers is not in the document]
johnpaliotta: Original issue without the --external:

Activating extension 'vectorgroup.vectorcasttestexplorer' failed: Cannot find module './impl/format' Require stack: - c:\RDS\vector-vscode-vcast\out\extension.js - c:\Program Files\Microsoft VS Code\resources\app\out\vs\loader.js - c:\Program Files\Microsoft VS Code\resources\app\out\bootstrap-amd.js - c:\Program Files\Microsoft VS Code\resources\app\out\bootstrap-fork.js - .

--external is the wrong solution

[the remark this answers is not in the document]
ojcole: This happens because of how microsoft write their packages which unfortunately does not make them easily compatible with bundlers like esbuild. When writing CommonJS packages they tend to wrap things in such a way that the original require function is called indirectly and not directly which means in the source the bundler is not aware it is a real call to require. As such, we would need to provide the bundler with more context to get around this. Thankfully, the ESM module that is provided is written as you'd expect and we can tell esbuild to prefer using the ESM module by specifying `--main-fields=module,main` after the `esbuild` command. This is only an issue with `extension.js` building so I recommend changing
```
"esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/extension.js --external:vscode --format=cjs --platform=node",
```
to
```
"esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/extension.js --external:vscode --format=cjs --platform=node --main-fields=module,main",
```
This is not a perfect solution and it has potential to break things in our dependencies do not conform properly to supporting ESM according to sources, but I have tested it here and it seems to work.

References:
- https://stackoverflow.com/questions/69208729/how-to-get-esbuild-to-include-internal-dependencies-when-using-typescript
- https://github.com/evanw/esbuild/issues/1619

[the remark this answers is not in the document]
johnpaliotta: Thanks @ojcole!

src/utilities.ts
```diff
@@ -152,93 +157,131 @@ export function getChecksumCommand() {
...
-      vectorJSON["files.exclude"]
-    );
-    fs.writeFileSync(filePath, JSON.stringify(existingJSON, null, "\t"));
+      vectorJSON["files.exclude"]);
+    const jsoncEdits = 
+      jsonc.modify(
+        existingJSONasString, 
+        [filesExcludeString], 
+        mergedExcludeList,
+        { formattingOptions: { tabSize: 4, insertSpaces: true } });
```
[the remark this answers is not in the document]
johnpaliotta: Done!

aytey:
@Zbigor can we modify whatever tests we have no for debugging preparation to have an input launch.json _with_ a comment and check the comment is preserved after we do debug prep?

src/utilities.ts
```diff
@@ -152,93 +158,133 @@ export function getChecksumCommand() {
...
-  } catch {
-    // if file is empty ...
-    existingJSON = { configurations: [] };
+  let returnValue:jsonDataType|undefined = undefined;
+  
+  // Requires json-c parsing to handle comments etc.
+  const existingContents = fs.readFileSync (jsonPath).toString();
+  var parseErrors: jsonc.ParseError[] = [];  // not using programatically, for debug only
+  // note that jsonc.parse returns "real json" without the comments
+  const existingJSONdata = jsonc.parse(existingContents, parseErrors, jsoncParseOptions);
```
aytey, line 177: Should we make this into a `parseJsonCFile` function? This would avoid the duplication of code such as `var parseErrors` (which happens 3 times).

src/utilities.ts
```diff
@@ -152,93 +158,133 @@ export function getChecksumCommand() {
...
   );
 
-  const vectorConfiguration = vectorJSON.configurations[0];
+  // if we have a well formatted launch file with an array of configurations ...
+  if (existingLaunchData && existingLaunchData.jsonData.configurations && existingLaunchData.jsonData.configurations.length > 0) {
 
-  // now loop through launch.json to make sure it does not already have the vector config
-  let existingConfigFound = false;
+      // Remember that the vectorJSON data has the "configurations" level which is an array
+    const vectorConfiguration = vectorJSON.configurations[0];
```
aytey, line 202: This surprises me; so the JSON object returned by json-c parser is of a different format than the normal parser?

What happens if there are no configurations in the fie?

src/testPane.ts
```diff
@@ -571,9 +571,9 @@ function launchConfigExists(launchJsonPath: string): boolean {
   // this will check if launch.json has
   // the "VectorCAST Harness Debug" configuration defined
   let returnValue = false;
-  const existingJSON: any = loadLaunchFile(launchJsonPath);
-  if (existingJSON.configurations) {
-    for (const existingConfig of existingJSON.configurations) {
+  const existingJSONdata: any = loadLaunchFile(launchJsonPath);
+  if (existingJSONdata) {
```
aytey, line 575: What if `existingJSONdata.jsonData` doesn't have `configurations`?

Should it be: `if (existingJSONdata and existingJSONdata.jsonData.configurations)`?

src/utilities.ts
```diff
@@ -152,93 +158,133 @@ export function getChecksumCommand() {
   return globalCheckSumCommand;
 }
 
-export function loadLaunchFile(jsonPath: string): any {
+export interface jsonDataType {
+  jsonData:any;
```
aytey, line 162: Commented: I wanted this not to be `any`, but `jsonc.parse` _does_ return an `any` 😞

src/utilities.ts
```diff
@@ -152,93 +158,133 @@ export function getChecksumCommand() {
...
+        );
+        needToAddVectorLaunchConfig = false;
         break;
       }
     }
-  else {
-    // if file does not have "configuration" section ...
-    existingJSON.configurations = [];
+    if (needToAddVectorLaunchConfig) {
+      const whereToInsert = existingLaunchData.jsonData.configurations.length;
```
aytey, line 217: Again: what happens if `jsonData` _does not_ have `configurations`

src/vcastUtilities.ts
```diff
@@ -159,49 +162,71 @@ export function addIncludePath (fileUri: vscode.Uri) {
...
-    vscode.window.showErrorMessage(`Exception parsing {configurationFile} file, no changes made.  Check for syntax errors.  `);
+  let existingJSONasString: string;
+  
+  // Requires json-c parsing to handle comments etc.
+  existingJSONasString  = fs.readFileSync (fileUri.fsPath).toString();
+  var parseErrors: jsonc.ParseError[] = [];  // not using programatically, for debug only
+  // note that jsonc.parse returns "real json" without the comments
+  existingJSON = jsonc.parse(existingJSONasString, parseErrors, jsoncParseOptions);
+
+  if (existingJSON && existingJSON.configurations && existingJSON.configurations.length > 0) {
```
aytey, line 174: Okay, so there _is_ a chance we don't have `configurations`?

Can we please review this code to check the consistency of checking  `configurations`? We probably want `existingJSON && existingJSON.configurations && existingJSON.configurations.length > 0` everywhere (or a helper that encapsulates this)?

Zbigor:
> @Zbigor can we modify whatever tests we have no for debugging preparation to have an input launch.json _with_ a comment and check the comment is preserved after we do debug prep?

Done in PR #89 against this branch

src/utilities.ts
johnpaliotta: No the same format.  I just made the vector one a "complete" json file with the configuration level so that I can use it at line 247

src/testPane.ts
johnpaliotta: Good catch, thanks.   Fix pushed - see comment

src/utilities.ts
johnpaliotta: Already protected by check at line 213

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "vectorcasttestexplorer",
   "displayName": "VectorCAST Test Explorer",
   "description": "VectorCAST Test Explorer for VS Code",
-  "version": "1.0.12",
```
aytey, line 5: This is wrong.

src/testPane.ts
```diff
@@ -719,11 +719,16 @@ async function debugNode(
 
         createEmptyLaunchConfigFile(ourWorkspace, launchJsonPath);
         addLaunchConfiguration(launchJsonUri);
-      } else {
+      } 
+      else {
```
aytey, line 723: Are we formatting this file at all?

src/utilities.ts
aytey: Sorry, this is my bad; I thought this was `existingLaunchData`.

src/vcastUtilities.ts
```diff
@@ -159,49 +162,71 @@ export function addIncludePath (fileUri: vscode.Uri) {
...
+    vscode.window.showErrorMessage(statusMessages.join ("\n"));
     return;
   }
-
+  
   // when we get here we should always have a configurations array
-  // but we might now have an includePath, so add it if its missing
+  // but we might not have an includePath, so add it if its missing
   
   let configName = existingJSON.configurations[0].name;
```
aytey, line 187: This was in the old code, but why are we only doing it on `configurations[0]`?

johnpaliotta: yes same as before, we could improve to prompt the user for the configuration they are using as a separate issue

src/testPane.ts
johnpaliotta: Have not, happy to run black on it as a separate branch

aytey: Okay, cool.

package.json
johnpaliotta: The current release is 10 so 11 will be next.

src/vcastUtilities.ts
aytey: Okay!

package.json
aytey: Oh, who moved it to 12?!

aytey: Ah, https://github.com/vectorgrp/vector-vscode-vcast/pull/86

src/utilities.ts
johnpaliotta: I think that is over-functioning :)  maybe we can just make the parseErrors global since they are not used??

aytey: The prototype of `parse` has `errors?: {error: ParseErrorCode;}[]` -- can we try passing in `null`?

If we were doing something with `parseErrors`, then fine, but we're not currently.

aytey (approved):
All good from my end; we can do https://github.com/vectorgrp/vector-vscode-vcast/pull/84/files#r1542680704 if we want, but it doesn't block merging this if we're not fussed.

PR #86, opened 2024-03-27
Build option improvements

johnpaliotta:
Made the following changes:

- Separated options into two sections: General and Build Environment
- Moved the Enable Coded Test option to the Build section
- Added a new option for coverage kind to the Build section
  - Previously the coverage kind was hard-coded to statement
- Changed the environment rebuild process to incorporate changes to the two build options
  -  Previously the Re-Build command kept everything the same

### DIFF
--- a/python/vTestInterface.py
+++ b/python/vTestInterface.py
@@ -637,6 +637,9 @@ def updatedEnvironment (enviroPath, jsonOptions):
         returnString += "\n" + "-"*100 + "\n"
         returnString += cleanClicastStdOut (stdOutput)
 
+        os.remove (tempEnviroScript)
+        os.remove (tempTestScript)
+
         return returnString
 
 
--- a/python/vTestInterface.py
+++ b/python/vTestInterface.py
@@ -529,7 +529,6 @@ def getCodeBasedTestNames (filePath):
     return returnObject
 
 
-
 class testID:
     def __init__(self, enviroPath, testIDString):
         self.enviroName, restOfString = testIDString.split("|")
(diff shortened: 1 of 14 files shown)


python/vTestInterface.py
```diff
@@ -560,7 +580,99 @@ def validateClicastCommand (command, mode):
...
+
+                # write the orignal or updated line
+                enviroFile.write (whatToWrite)
+
+        # Finally delelete and re-build the environment using the updated script
+        # and load the existing tests -> which duplicates what enviro rebuild does.
+        with open(commandFileName, "w") as commandFile:
+            # Improvement needed: vcast bug: 100924
+            shutil.rmtree (enviroName)
+            #commandFile.write(f"-e{enviroName} enviro delete\n")
```
Zbigor, line 640: I guess the `delete` command doesn't work properly, so we are using `shutil.rmtree`

python/vTestInterface.py
```diff
@@ -560,7 +580,99 @@ def validateClicastCommand (command, mode):
             raise UsageError()
         
 
-def processCommand (mode, clicast, pathToUse, testString="") -> dict:
+def updatedEnvironment (enviroPath, jsonOptions):
+    """
+    pathToUse is the full path to the environment directory
+    jsonOptions has the new values of ENVIRO.* commands for the enviro script
+    e.g.  ENVIRO.COVERAGE_TYPE: Statement
+
```
Zbigor, line 588: We are not editing the original `.env` at any point, it might make sense to point this out in the function description here

Zbigor (approved):
Changes look good, left only a couple of minor comments. I made a branch with minor changes to improve e2e stability off of this one (PR #87 ), e2e tests are all good

PR #89, opened 2024-03-28
Adapted one of the debug prep e2e test to add a comment to the generated json config

Zbigor:


johnpaliotta:
Should we check that the comment is maintained after our extension does some work on the file?

Zbigor:
> Should we check that the comment is maintained after our extension does some work on the file?

I'll add a check that the comment is still in there after the debug prep is done

Zbigor:
> Should we check that the comment is maintained after our extension does some work on the file?

Done

PR #97, opened 2024-04-05
Rearchitecture of TypeScript files

johnpaliotta:
Purpose is to organize the commands and calls to clicast, vpython, and vcastqt in a logical way.
This was a pre-requisite for implementing an environment server.

### DIFF
--- a/python/clicastInterface.py
+++ b/python/clicastInterface.py
@@ -1,0 +1,182 @@
+import os
+import re
+import shutil
+import subprocess
+import sys
+import time
+
+from vector.lib.core.system import cd
+
+
+commandFileName = "commands.cmd"
+globalClicastCommand = ""
+
+
+enviroNameRegex = "-e([^\s]*)"
+
+
+def getEnviroPathFromCommand(command):
+
+    # TBD in the future we will change the caller to pass in the environment path
+    # No error handling because the caller will guarantee that we have a valid command
+    match = re.search(enviroNameRegex, command)
+    enviroName = match.group(1)
+    enviroPath = os.path.join(os.getcwd(), enviroName)
+
+    return enviroPath
+
+
+def runClicastCommandCommandLine(commandToRun):
+    """
+    A wrapper for the subprocess.run() function
+    """
+    try:
+        # note: shell=true, requires commandToRun to be a string
+        result = subprocess.run(
+            commandToRun,
+            check=True,
+            stdout=subprocess.PIPE,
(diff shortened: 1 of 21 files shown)


src/testPane.ts
```diff
@@ -314,12 +305,11 @@ function getEnvironmentList(baseDirectory: string): string[] {
...
 
 // These variables are used in the package.json to control the display of context menu items
-// Search for 'vectorcastTestExplorer.vcastEnviroList | vcastHasCodedTestsList' in package.json 
+// Search for 'vectorcastTestExplorer.vcastEnviroList | vcastHasCodedTestsList' in package.json
 // to see where we reference them
 // this list in a "when" clause
-export var vcastEnviroList: string[] = [];
-export var vcastHasCodedTestsList: string[] = [];
-
+let vcastEnviroList: string[] = [];
```
Zbigor, line 311: no need for export anymore (we're not using these in other files)?

src/vcastAdapter.ts
```diff
@@ -0,0 +1,326 @@
...
+
+const path = require("path");
+
+// ------------------------------------------------------------------------------------
+// Direct clicast Calls
+// ------------------------------------------------------------------------------------
+
+// Check License
+export function vcastLicenseOK(): boolean {
+  const commandToRun = `${clicastCommandToUse} tools has_license`;
```
Zbigor, line 42: Should we use clicast interface?

Zbigor (approved):
After going through the diffs, I mapped out where things got moved and how they changed. Changes look good

python/clicastInterface.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+enviroNameRegex = "-e([^\s]*)"
+
+
+def getEnviroPathFromCommand(command):
+
+    # TBD in the future we will change the caller to pass in the environment path
+    # No error handling because the caller will guarantee that we have a valid command
+    match = re.search(enviroNameRegex, command)
```
aytey, line 22: Who generates what's passed to this? It is valid to do `-e<space><env name>`.

The regex could be:

```py
enviroNameRegex = "-e\s*([^\s]*)"
```

src/testPane.ts
johnpaliotta: The export was removed in this change :)

johnpaliotta:
> After going through the diffs, I mapped out where things got moved and how they changed. Changes look good

Thanks!

Issue #96, opened 2024-04-05
shouldPromptForIncludePath function should expand environment variables


matefriedl:
`"${env:VECTORCAST_DIR}/vunit/include"` is a reasonable setting for the `includePath` option of the `c_cpp_properties.json` file. The extension considers this wrong and prompts to replace it with an absolute path. This notification has to be dismissed every time one reloads/opens the editor.

![image](https://github.com/vectorgrp/vector-vscode-vcast/assets/63063644/f68c054e-30c6-4057-8dc6-89d01985f90d)

aytey:
I had a chat with @matefriedl on this.

A few points:

- The "pain point" here is that the extension *does not recognise* that `"${env:VECTORCAST_DIR}/vunit/include"` is a valid option (so it keeps complaining)
- The request isn't (really) to _generate_ `"${env:VECTORCAST_DIR}/vunit/include"`, but just not to complain about it when it is there
    - This is "complicated" because not all customers will have `VECTORCAST_DIR`
- If do the "update", the extension does _remove_ `"${env:VECTORCAST_DIR}/vunit/include"` and replaces it with `/path/to/fully/qualified/vunit/include`, because the extension thinks that `"${env:VECTORCAST_DIR}/vunit/include"` is a path to an out-dated install

PR #102, opened 2024-04-29
Fix for issue: #96, extension does not recognize vunit include path that uses an environment variable

johnpaliotta:


johnpaliotta:
@matefriedl could you try this update?

johnpaliotta:
@aytey ok to merge?

PR #104, opened 2024-04-30
Updated use of dataAPI to retrieve coverage information

johnpaliotta:
The extension was previously using an "unsupported VectorCAST API", so this moves to the officially supported API.

### DIFF
--- a/python/vTestInterface.py
+++ b/python/vTestInterface.py
@@ -309,12 +309,9 @@ def getCoverageData(sourceObject):
         if os.path.exists(sourceObject.path):
             for line in sourceObject.iterate_coverage():
                 metrics = line.metrics
-                if (
-                    metrics.max_covered_statements == 1
-                    or metrics.annotations_statements == 1
-                ):
+                if metrics.has_any_coverage():
                     coveredString += str(line.line_number) + ","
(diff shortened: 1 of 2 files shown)


python/vTestInterface.py
```diff
@@ -252,62 +251,32 @@ def getTestDataVCAST(enviroPath):
...
-        unitInfo = dict()
-        unitInfo["path"] = sourcePath
-        unitInfo["functionList"] = getFunctionData(sourceObject)
-        unitInfo["cmcChecksum"] = checksum
-        unitInfo["covered"] = covered
-        unitInfo["uncovered"] = uncovered
-        unitList.append(unitInfo)
-
-    capi.close()
+        if sourceObject.is_instrumented:
```
aytey, line 268: Did `capi.SourceFile` previously only give instrumented files?

I am unsure of the implication of this change if, e.g., you have an environment with `COVERAGE: NONE` -- does that affect `unitInfo["functionList"] = getFunctionData(sourceObject)`?

aytey:
E2E tests pass. Happy to merge when the comment is addressed.

aytey (approved):

PR #105, opened 2024-05-02
Add support for coded mocks

johnpaliotta:
This PR adds support to VectorCAST extensions for coded mocks.

Importantly, it adds two new auto-complete features:

* `// vmock` (followed by a unit and then a subprogram)
* `vsession`


python/tstUtilities.py
```diff
@@ -466,7 +470,420 @@ def splitExistingLine(line):
...
+            if openParen == 0:
+                break
+        elif char == "(":
+            openParen += 1
+
+    # one spcial case, for const functions the return type
+    # will have const appended, so strip this.
+    returnType = parameterizedName[index + 1 :]
+    if returnType.endswith("const"):
+        returnType = returnType[:-6]
```
aytey, line 711: Please use `-len("const")` and not `-6`.

aytey: oh, you probably want `len(" const")` (including the space)

python/tstUtilities.py
```diff
@@ -609,11 +609,44 @@ def getFunctionSignature(api, functionObject):
...
+
 def functionIsOperator(functionName):
     """
     So that we have one place to adjust if we find bugs :)
     """
-    return "::operator" in functionName or functionName.startswith("operator")
+    returnValue = False
+    if "::operator" in functionName:
+        operatorIndex = functionName.find("::operator")
+        returnValue = isOperator(functionName[operatorIndex + len("::operator") :])
```
aytey, line 643: Like this can be:

`returnValue = isOperator(functionName.split("::operator") + "::operator")`

which avoids the explicit find + list splice.

PR #107, opened 2024-05-02
Fix for issue: 106, relative paths for coded tests

johnpaliotta:
The only way to get relative paths is to edit the .tst so this is a low priority issue.


johnpaliotta:
@Zbigor   please review

python/vTestInterface.py
```diff
@@ -136,7 +136,7 @@ def getPassFailString(test):
     return XofYString(numerator, denominator)
 
 
-def generateTestInfo(test):
+def generateTestInfo(enviroPath, test):
```
aytey, line 139: Need to update the docstring

python/vTestInterface.py
```diff
@@ -155,8 +155,10 @@ def generateTestInfo(test):
     if vpythonHasCodedTestSupport and test.coded_tests_file:
         # guard against the case where the coded test file has been renamed or deleted
         # or dataAPI has a bad line nuumber for the test, and return None in this case.
-        if os.path.exists(test.coded_tests_file) and test.coded_tests_line > 0:
-            testInfo["codedTestFile"] = test.coded_tests_file
+        enclosingDirectory = os.path.dirname (enviroPath)
```
aytey, line 158: Please run `black`

aytey:
@johnpaliotta: do we need a test for this? How do we get a relative file path for a coded test?

python/vTestInterface.py
johnpaliotta: > Please run `black`

done

python/vTestInterface.py
johnpaliotta: > Need to update the docstring

Done

johnpaliotta:
> @johnpaliotta: do we need a test for this? How do we get a relative file path for a coded test?

Probably should have a test.
As I said in the first comment above :)  ... you have to manually edit the .tst file.

PR #111, opened 2024-05-24
Github Actions Pipeline for building and testing the extension

Zbigor:
Together with @SalvatoreZagaria 
- Both unit tests and e2e tests are run inside docker containers
- E2E Tests are run in parallel for vc23 and vc24
  - vc24 E2E tests take longer since they include the tests for coded test workflows
- E2E tests now stop on failure
- Unit test coverage info is uploaded to and can be viewed on `codecov.io`
- Github Actions job summary includes download links for 
  - vsix file
  - coverage info
  - error screenshots 
- E2E Error messages and their locations are parsed from the logs and are shown in the summary


.vscode/launch.json
```diff
@@ -26,11 +26,45 @@
...
+      "console": "integratedTerminal"
+    },
+    {
+      "name": "getCompletion",
+      "type": "python",
+      "request": "launch",
+      "program": "${workspaceFolder}/python/testEditorInterface.py",
+      "cwd": "C:/RDS/VectorCAST/testExplorer/unitTests",
+      "env": {
+        "PYTHONPATH": "C:/vcast/vc23/python"
```
aytey, line 59: What's with these changes?

src/vcastTestInterface.ts
```diff
@@ -902,4 +902,4 @@ export async function openCodedTest(testNode: testNodeType) {
   if (fs.existsSync(testNode.testFile)) {
     openFileWithLineSelected(testNode.testFile, testNode.testStartLine - 1);
   }
-}
+}
```
aytey, line 905: I think we can revert this?

.vscode/launch.json
Zbigor: These were removed in ea4948a8e6419375b1d8b874586b4a4f313eca86 . They ended up here when I was branching from a stable version and dealing with merge conflicts. As they seemed to be unused commands now, I can get rid of these

src/vcastTestInterface.ts
Zbigor: This a new line at the end of file, but yeah, we can revert if you want

python/vTestInterface.py
```diff
@@ -588,3 +588,4 @@ def main():
         returnCode = 1
 
     sys.exit(returnCode)
+
```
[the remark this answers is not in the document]
Zbigor: same as above

src/vcastTestInterface.ts
aytey: Just to make it cleaner on the diff.

Then we can be sure there's no "application" changes in this.

python/vTestInterface.py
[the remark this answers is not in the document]
aytey: Just to make it cleaner on the diff.

Then we can be sure there's no "application" changes in this.

.vscode/launch.json
aytey: Ah, yeah, would be good to remove these

PR #123, opened 2024-06-19
Several usability improvements

johnpaliotta:
1. Sorted "Coded Tests" to the top of the list of testable functions for each unit in the test pane.
2. Reduced the number of file checksums computed for code coverage purposes.
3. Fixed issue: #128


src/vcastTestInterface.ts
```diff
@@ -57,8 +57,28 @@ const path = require("path");
 
 export const vcastEnviroFile = "UNITDATA.VCD";
 
+// Creating a cache for the checksums so we don't constantly re-run the command
+interface checksumCacheType {
```
aytey, line 61: Sonar doesn't like this: https://rules.sonarsource.com/typescript/type/Code%20Smell/RSPEC-101/

python/clicastInterface.py
```diff
@@ -201,9 +201,11 @@ def rebuildEnvironment(enviroPath, jsonOptions):
         rebuildEnvironmentUsingClicastReBuild(enviroPath)
 
 
+codeTestCompileErrorCode=98
 def executeTest(testIDObject):
     # since we are doing a direct call to clicast, we need to quote the parameters
     # separate variable because in the future there will be additional parameters
+    global codeTestCompileErrorCode
```
aytey, line 208: You're not assigning to this, so it doesn't need to be marked as `global`.

python/vTestInterface.py
```diff
@@ -391,7 +392,9 @@ def executeVCtest(enviroPath, testIDObject, generateReport):
         returnText = ""
 
         returnCode, commandOutput = clicastInterface.executeTest(testIDObject)
-        if generateReport:
+        # return code 98 means the the coded test failed to compile so
```
aytey, line 395: Should this comment not be against the variable declaration rather than its usage?

src/testPane.ts
```diff
@@ -173,7 +173,7 @@ function addTestNodes(
 }
 
 const codedTestFunctionName = "coded_tests_driver";
-const codedTestDisplayName = "Coded Tests";
+const codedTestDisplayName = " Coded Tests";
```
aytey, line 176: Is this leading space just so it goes to the top of the list?

src/vcastTestInterface.ts
```diff
@@ -75,14 +95,20 @@ function getChecksum(filePath: string) {
...
-    // this will crash if something is wrong with the result
+    // this will throw if something is wrong with the result
     try {
       if (commandOutputString.includes("ACTUAL-DATA")) {
         const pieces = commandOutputString.split("ACTUAL-DATA", 2);
         returnValue = Number(pieces[1].trim());
       } else {
         returnValue = Number(commandOutputString);
       }
+      // only save intot the cache if we get a valid checksum
```
aytey, line 106: intot -> into

python/clicastInterface.py
```diff
@@ -201,9 +201,11 @@ def rebuildEnvironment(enviroPath, jsonOptions):
         rebuildEnvironmentUsingClicastReBuild(enviroPath)
 
 
+codeTestCompileErrorCode=98
```
aytey, line 204: Actions failed because of a `black` diff:

```patch
diff --git a/python/clicastInterface.py b/python/clicastInterface.py
index 8cc80e9..4d9f10b 100644
--- a/python/clicastInterface.py
+++ b/python/clicastInterface.py
@@ -201,7 +201,9 @@ def rebuildEnvironment(enviroPath, jsonOptions):
         rebuildEnvironmentUsingClicastReBuild(enviroPath)


-codeTestCompileErrorCode=98
+codeTestCompileErrorCode = 98
+
+
 def executeTest(testIDObject):
     # since we are doing a direct call to clicast, we need to quote the parameters
     # separate variable because in the future there will be additional parameters
```

.vscode/launch.json
```diff
@@ -19,15 +19,15 @@
       "type": "python",
       "request": "launch",
       "program": "${workspaceFolder}/python/vTestInterface.py",
-      "cwd": "C:/RDS/VectorCAST/FOO",
+      "cwd": "C:/RDS/VectorCAST/CodeBasedTests/unitTests",
```
aytey, line 22: @johnpaliotta happy to discuss, but I think we should start rejecting PRs that change this file.

This feel seems to get changed in every PR (I understand why), but I think we should start not including changes to this file inside of "functional" changes to the extension.

johnpaliotta: Let's just remove it from the repo

src/testPane.ts
johnpaliotta: yes

src/vcastTestInterface.ts
johnpaliotta: what does it want Checksum *?

.vscode/launch.json
aytey: How about we clean it up, and move it to `.vscode/launch.json.example`? Then you can copy/make new copies as you like, but they won't be tracked.

To be clear: I see value in this file being in the repo, but I'm trying to avoid the "noise".

src/vcastTestInterface.ts
aytey: Yeah

.vscode/launch.json
johnpaliotta: Sure, it saves me a ton of time to have it match my setup

aytey (approved):

PR #127, opened 2024-06-19
Improvements to GitHub Actions

SalvatoreZagaria:
- The E2E tests were already ending when a failure happened, but GHA was not detecting the failure. We now parse the output to see if the suite has failed, and in case, we return a 1 status code.
- The runs naming has been changed to "branch-name is being tested"
- We have a new job that checks the python scripts format with black
- We don't run the jobs if the trigger comes from a draft PR
- We upload screenshots separately (not in an archive)
- Removed container's option ` -v /var/run/dbus/system_bus_socket:/var/run/dbus/system_bus_socket:rw`, suspected to be the reason why the containers hang when running E2E tests at the same time


Zbigor (approved):
Changes and the summary look good

PR #132, opened 2024-06-20
Coded mock cleanup

aytey:
The main purpose of this branch is to centralise the logic for identifying "functions of interest".

This now makes `getFunctionList` be the central point that both .tst editing and coded-test editing uses to identify functions that should be tested/accessed/etc.

### DIFF
--- a/python/tstUtilities.py
+++ b/python/tstUtilities.py
@@ -18,7 +18,9 @@ globalOutputLog = list()
 def getNameDictFromObjectList(objectList):
     """
     This will take a list of dataAPI objects and return
-    a list of names ... note all objects have name attributes
+    a dictionary of names to objects.
+
+    Note all objects have name attributes.
     """
     returnDict = {}
     for object in objectList:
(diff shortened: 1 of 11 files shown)


python/tstUtilities.py
```diff
@@ -814,24 +821,20 @@ def processVMockDefinition(enviroName, lineSoFar):
...
 
-        whatToReturn = generateVMockDefitionForUnitAndFunction(
+        whatToReturn = generateVMockDefinitionForUnitAndFunction(
             unitObject, functionObject
         )
 
         returnData.choiceKind = choiceKindType.Snippet
         returnData.choiceList.append(whatToReturn)
 
-    api.close()
```
aytey, line 828: You don't need this, `api` will get closed when it goes out of scope

python/tstUtilities.py
```diff
@@ -893,7 +896,6 @@ def processTstLine(enviroName, line):
         else:
             returnData = processStandardLines(api, pieces, triggerCharacter)
 
-        api.close()
```
aytey, line 896: You don't need this, `api` will get closed when it goes out of scope

PR #143, opened 2024-06-26
Strip out another case where the vcast version miss-match error is shown in the output pane

johnpaliotta:


src/vcastCommandRunner.ts
```diff
@@ -62,7 +62,9 @@ function processExceptionFromExecuteCommand(
 
   // 99 is a warning, like a mismatch opening the environment
   if (error && error.status == 99) {
-    commandStatus.stdout = error.stdout.toString();
+    let stdoutString =  error.stdout.toString();
+    // Needed because of the annoying version miss-match message from vcast
+    commandStatus.stdout = stdoutString.split ("ACTUAL-DATA", 2)[1].trim ()
```
aytey, line 67: In the past, when I've "broken" things, we get the Python traceback in the message area.

Is this going to break that?

johnpaliotta: The py utilities ALWAYS "print" ACTUAL-DATA as the first thing in the main to make the processing in TS easy.
So any crash etc will be after the ACTUAL*

aytey: `testEditorInterface.py` prints it _after_ doing its processing

johnpaliotta: Good point we can  move them to global scope or the start of the main for consistency

aytey: Start of main would match what's in `vTestInterface.py`

johnpaliotta: But not crc32.py :)  ... done

aytey, line 67: `prettier` wants to change this, *but* there's quite a few `prettier` changes against `src`.

I suggest we merge this and then do a `prettier` branch.

aytey: I prettier'd this one file.

aytey (approved):

PR #144, opened 2024-06-28
Gha e2e jobs parallelisation

SalvatoreZagaria:
We are dividing the e2e tests specs into independent groups, allowing the Github Actions workflow to retrieve all the groups and run several combinations of vcastVersion-group in parallel, saving a big amount of time.

### DIFF
--- a/.github/workflows/run-tests-workflow.yml
+++ b/.github/workflows/run-tests-workflow.yml
@@ -368,5 +368,5 @@ jobs:
       - name: Black
         shell: bash
         run: |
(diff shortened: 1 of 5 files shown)


Zbigor (changes requested):
Looks good. I think it'd make sense to update `tests/internal/e2e/README.md`.
- We need to set some environment variables now to be able to run anything
- We should describe how/where to add new specs/specs groups after this change

Zbigor (approved):

PR #145, opened 2024-07-02
Adding pre-commit hook that runs black before commit

SalvatoreZagaria:


.githooks/install_hooks.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+ROOT=$(dirname "$(realpath "$0")")
+cd $ROOT/..
+git config core.hooksPath .githooks
```
aytey, line 5: I wonder if we need the `cd`? Wouldn't it be enough to just do `git config core.hooksPath ${ROOT}.githooks`?

SalvatoreZagaria: I've done it just to make sure that we are in the repository when we run this command, and not somewhere else

aytey (approved):

PR #150, opened 2024-07-11
Start working on improved tests with mocked types from DataAPI

aytey:


aytey:
I'm WIP on a script that generates `coded_mock_test_data.py` from a set of environments

aytey:
We have no namespace tests yet (need an example where we have something that looks like a class, but is actually a namespace)

aytey:
I need to clean-up the test data; I removed it for now.

aytey:
I don't think we want this now.

PR #151, opened 2024-07-11
Run tsc before running vitest

aytey:
Tests are now type-checked with `tsc` before running `vitest`.

In the coded_mock branch, we would have received:

```
tests/unit/utils.ts:135:31 - error TS2345: Argument of type 'TextDocuments' is not assignable to parameter of type 'TextDocument'.
  Type 'TextDocuments' is missing the following properties from type 'TextDocument': uri, languageId, version, getText, and 3 more.

135   return getTstCompletionData(documents, completion);
```

and the tests wouldn't have run.

### DIFF
--- a/package.json
+++ b/package.json
@@ -603,7 +603,7 @@
     "esbuild": "npm run esbuild-base -- --sourcemap && npm run esbuild-server -- --sourcemap",
     "package": "vsce package",
     "publish": "vsce publish",
-    "test": "vitest run --reporter verbose --coverage",
+    "test": "tsc --project tsconfig_test.json && vitest run --reporter verbose --coverage",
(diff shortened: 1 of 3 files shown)


aytey:
@SalvatoreZagaria: if `npm test` fails, what happens to GHA?

aytey:
> @SalvatoreZagaria: if `npm test` fails, what happens to GHA?
> 
> To clarify: this isn't `vitest` failing, this `npm test` giving a non-zero return code -- does this fail the tests when viewed by GHA? If not, can we make that happen?

I guess it will fail, based on https://github.com/vectorgrp/vector-vscode-vcast/pull/153

johnpaliotta (approved):

PR #153, opened 2024-07-11
Run tsc before package

aytey:
This PR adds `tsc` to be run before `npm run package` (and `vsce package` should no longer be used directly).


aytey:
This PR depends on https://github.com/vectorgrp/vector-vscode-vcast/pull/152

PR #157, opened 2024-07-19
Use updated API

aytey:


python/dataAPIutilities.py
```diff
@@ -22,12 +22,7 @@ def getParameterList(functionObject):
         if parameterObject.name != "return":
             paramIndex += 1
             # get parameterObject.orig_declaration
-            declarationToUse = getOriginalDeclaration(parameterObject)
```
johnpaliotta, line 25: Why don't I see the removal of the function definition getOriginalDeclaration in the diff?

python/dataAPIutilities.py
```diff
@@ -155,31 +161,42 @@ def getInstantiatingClass(api, functionObject):
     # Andrew please add exactly what we want to the Confluence page
```
johnpaliotta, line 161: Update comment with issue# here

python/dataAPIutilities.py
```diff
@@ -155,41 +161,50 @@ def getInstantiatingClass(api, functionObject):
...
     return instantiatingClass
 
 
-# ----------------------------------------------------------------------------------------
-# PCT-FIX-NEEDED - would be nice if Ian would generate the apply functions for us
-# so that we don't have to do any of the processing below these lines
-# ----------------------------------------------------------------------------------------
+def getFunctionNameForAddress(api, functionObject):
+    # PCT-FIX-NEEDED - 
+    # Andrew - Add a descriptio of what we want an an issue#
```
johnpaliotta, line 205: Add a description of what we want an issue number

python/dataAPIutilities.py
aytey: Oversight; it got forgotten when I changed from use_generate_mock_declaration.

python/dataAPIutilities.py
aytey: This is already there (including a detailed example), but you haven't given it an issue number (because it isn't a "bug fix" but a "feature request").

Do you want me to add a bug number to the Confluence page?

python/dataAPIutilities.py
aytey: This is already there (including a detailed example), but you haven't given it an issue number (because it isn't a "bug fix" but a "feature request").

Do you want me to add a bug number to the Confluence page?

PR #160, opened 2024-07-23
Bump prettier to 3.3.3

aytey:


package-lock.json
```diff
@@ -10410,20 +10409,6 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
-    "node_modules/xo/node_modules/prettier": {
```
aytey, line 10413: I don't know where this has gone or if it is a problem.

Zbigor: I guess xo "installs it" and "gets an override. If we agree to use `npx prettier -w` for now, it should be fine. Whenever we actually get to using `xo`, we'll define everything there, including `eslint` config etc.

aytey: I've asked @SalvatoreZagaria to add a new check that does `npx prettier --check $(echo **/*.ts | tr ' ' '\n' | grep -v "node_modules" | xargs)`.

SalvatoreZagaria: https://github.com/vectorgrp/vector-vscode-vcast/pull/161

PR #161, opened 2024-07-23
Adding `prettier` check to GHA

SalvatoreZagaria:
Add `prettier` check to the GHA workflow. Both `black` an `prettier` jobs fail in case of unformatted files.

### DIFF
--- a/.github/workflows/run-tests-workflow.yml
+++ b/.github/workflows/run-tests-workflow.yml
@@ -348,7 +348,10 @@ jobs:
       - name: Exit with code
         shell: bash
         if: always()
-        run: exit $(cat exitcode.txt)
+        run: |
+          if [ -f exitcode.txt ]; then
+              exit $(cat exitcode.txt)
+          fi
 
   prettier:
     if: github.event.pull_request.draft == false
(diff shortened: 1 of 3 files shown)


docker/Ubuntu24-Dockerfile-CI
```diff
@@ -44,7 +44,8 @@ RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - && \
     npm install -g typescript && \
     npm install -g esbuild && \
     npm install -g xo && \
-    npm install -g tsx
+    npm install -g tsx && \
+    npm install -g prettier
```
aytey, line 48: We shouldn't need this, as it will be installed by the project

aytey: Another PR: but probably _all_ of these `npm install`s can go

.github/workflows/run-tests-workflow.yml
```diff
@@ -329,8 +329,55 @@ jobs:
...
+          exit 1
+
+      - name: Check out repository (retry)
+        if: failure()
+        uses: actions/checkout@v4
+
+      - name: Run prettier
+        shell: bash
+        run: |
+          npx prettier --check $(echo **/*.ts | tr ' ' '\n' | grep -v "node_modules" | xargs) > output.log 2>&1
```
aytey, line 375: Let's change this to be:

```
npx prettier -c .
```

and add a `.prettierignore`:

- https://stackoverflow.com/a/70715829

PR #164, opened 2024-07-29
Mock cleanup

aytey:
Some general clean-up using `prospector` and `vulture`.


python/dataAPIutilities.py
```diff
@@ -20,19 +17,25 @@ def functionCanBeMocked(functionObject):
     # these <<INIT>> functions should not be in the list
     # Waiting for PCT fix of FB: 101353.
     """
+
+    # function to not be shown in the functions list
```
johnpaliotta, line 21: Was global so that the list only gets built once not on every call

python/vmockGenerator.py
```diff
@@ -124,14 +122,17 @@ def generateAllVMockDefinitions(enviroPath):
 script_file = f"{basename}.tst"
 
 
-def generate_test_file(enviro_path, prepend=[]):
```
johnpaliotta, line 127: Defaulting to an empty list seems cleaner

johnpaliotta (approved):

PR #165, opened 2024-07-30
Added tests for uncovered lines in pythonUtilities.ts

Den1552:
## Summary

This PR adds new test cases for previously uncovered lines in `pythonUtilities.ts`.

## Changes

- Changed the import of `execSync` in `pythonUtilities.ts` to ensure `execSync` is properly mocked.
- Uploading `coverage-final.json` to `codecov` instead of `clover.xml`, to make sure we are uploading aggregated coverage results
## Additions

- `python-utilities.valid.test.ts`, which includes a test case for valid paths.
- `python-utilities.invalid.test.ts`, which includes a test case for invalid paths.

## Note

Since `process.argv` cannot be mocked in two test cases within the same file, there are two separate files to test `pythonUtilities.ts`.


Zbigor (approved):
lgtm

PR #166, opened 2024-08-05
Added tests for ctCompletion

Den1552:
## Summary

Fixed server tests and added tests for `ctCompletion` and `pythonUtilities` .

## Added
- Added test file for ctCompletion.ts.
- Added test files for pythonUtilities.ts.

## Changes

- Run `xo --fix `on files.
- Modified `generateCompletionData` to enable processing of coded tests.
- Modified `tst-completion.test.ts`, since `generateCompletionData` has an additional parameter.
- Modified `setup-unit-test.ts`, since cpp file needs to be created.
`

### DIFF
--- a/server/pythonUtilities.ts
+++ b/server/pythonUtilities.ts
@@ -1,10 +1,10 @@
 import fs = require("fs");
 import path = require("path");
-import { execSync } from "child_process";
-import { cleanVcastOutput } from "../src-common/commonUtilities";
+import { execSync } from "node:child_process";
+import { cleanVcastOutput } from "../src-common/commonUtilities.js";
 
-let testEditorScriptPath: string | undefined = undefined;
-let vPythonCommandToUse: string | undefined = undefined;
+let testEditorScriptPath: string | undefined;
+let vPythonCommandToUse: string | undefined;
 
 export function setPaths(
   _testEditorScriptPath: string,
@@ -18,16 +18,19 @@ export function updateVPythonCommand(newPath: string) {
   vPythonCommandToUse = newPath;
 }
 
+export function getVPythonCommand() {
+  return vPythonCommandToUse;
+}
+
 function initializeScriptPath() {
   // The client passes the extensionRoot and vpython command in the args to the server
   // see: client.ts:activateLanguageServerClient()
 
   const extensionRoot = process.argv[2];
-  // if we have not been sent the explicit path to use
+  // If we have not been sent the explicit path to use
   // fetch it from the command line arguments
-  if (vPythonCommandToUse == undefined) {
-    vPythonCommandToUse = process.argv[3];
-  }
+  vPythonCommandToUse ??= process.argv[3];
+
   const pathToTestEditorInterface = path.join(
     extensionRoot,
     "python",
@@ -50,11 +53,10 @@ export function runPythonScript(
   enviroName: string,
   payload: string
 ): any {
-  // this is currently not used as the actual server mode is unused
+  // This is currently not used as the actual server mode is unused
   if (testEditorScriptPath == undefined) {
     initializeScriptPath();
   }
-
   // As an alternative to using a server, we call vpython each time we need some data
 
   // NOTE: we cannot use executeCommand() here because it is in the client only!
@@ -63,7 +65,7 @@ export function runPythonScript(
   const commandToRun = `${vPythonCommandToUse} ${testEditorScriptPath} ${action} ${enviroName} "${payload}"`;
   const commandOutputBuffer = execSync(commandToRun).toString();
 
-  // two steps to make debugging easier
+  // Two steps to make debugging easier
   const outputString = cleanVcastOutput(commandOutputBuffer);
   return JSON.parse(outputString);
 }
@@ -78,9 +80,10 @@ export function getChoiceDataFromPython(
   lineSoFar: string
 ): any {
   const jsonData = runPythonScript(kind, enviroName, lineSoFar);
-  for (const msg of jsonData.messages) {
-    console.log(msg);
+  for (const message of jsonData.messages) {
+    console.log(message);
   }
+
   return jsonData;
 }
 
@@ -88,25 +91,27 @@ export function getHoverStringForRequirement(
   enviroName: string,
   requirementKey: string
 ): any {
-  let returnValue: string = "";
+  let returnValue = "";
   const jsonData = runPythonScript(
     "choiceList-tst",
     enviroName,
     "TEST.REQUIREMENT_KEY:"
   );
-  for (const msg of jsonData.messages) {
-    console.log(msg);
+  for (const message of jsonData.messages) {
+    console.log(message);
   }
+
   for (const line of jsonData.choiceList) {
     if (line.startsWith(requirementKey)) {
-      // raw data looks like:  <key> ||| <title> ||| <description>
+      // Raw data looks like:  <key> ||| <title> ||| <description>
       const pieces = line.split("|||");
-      // title often has double quotes in our examples so strip those too
-      const title = pieces[1].trim().replace(/['"]+/g, "");
+      // Title often has double quotes in our examples so strip those too
+      const title = pieces[1].trim().replaceAll(/['"]+/g, "");
       const description = pieces[2].trim();
       returnValue = `${title} \n\n ${description}`;
       break;
     }
   }
+
   return returnValue;
 }
--- a/tests/unit/ct-completion.test.ts
+++ b/tests/unit/ct-completion.test.ts
@@ -1,0 +1,155 @@
+import { describe, expect, test, vi, afterEach, beforeEach } from "vitest";
+import {
+  getCompletionPositionForLine,
+  generateCompletionData,
+} from "./utils.js";
+
+const unitTst = `
+// ---------------------------------------------------------------------------------------
+// Simple Example - new
+// vmock vmock_examples simpleFunction 
+int vmock_vmock_examples_simpleFunction(::vunit::CallCtx<> vunit_ctx, char param1, float param2) {
+  // Enable Stub: vmock_vmock_examples_simpleFunction_enable_disable(vmock_session);
+  // Disable Stub: vmock_vmock_examples_simpleFunction_enable_disable(vmock_session, false);
+
+  return 100;
+}
+void vmock_vmock_examples_simpleFunction_enable_disable(vunit::MockSession &vmock_session, bool enable = true) {
+    using vcast_mock_rtype = int ;
+    vcast_mock_rtype (*vcast_fn_ptr)(char param1, float param2)  = &simpleFunction;
+    vmock_session.mock <vcast_mock_rtype (*)(char param1, float param2)> ((vcast_mock_rtype (*)(char param1, float param2))vcast_fn_ptr).assign (enable ? &vmock_vmock_examples_simpleFunction : nullptr);
+}
+
+VTEST(vmockExamples, simpleTest2) {
+
+  auto vmock_session = ::vunit::MockSession();
+  vmock_vmock_examples_simpleFunction_enable_disable(vmock_session);
+  VASSERT_EQ (100, simpleFunction ('a', 1.0));
+
+  // disable the stub, which means the real code will return param1 'a' or 97
+  vmock_vmock_examples_simpleFunction_enable_disable(vmock_session, false);
+  VASSERT_EQ (97, simpleFunction ('a', 1.0));
+
+}
+`;
+
+const timeout = 30_000; // 30 seconds
+
+describe("Testing pythonUtilities (valid)", () => {
+  beforeEach(() => {
+    // To achieve 100% coverage
+    // --> Need to mock getChoiceDataFromPython, since "extraText" is not implemented yet
+    vi.mock("../../server/pythonUtilities", async () => {
+      const actual = await vi.importActual<
+        typeof import("../../server/pythonUtilities")
+      >("../../server/pythonUtilities");
+      return {
+        ...actual,
+        getChoiceDataFromPython: vi.fn(
+          (kind: string, enviroName: string, lineSoFar: string) => {
+            if (lineSoFar.includes("// vmock")) {
+              return {
+                choiceKind: "File",
+                choiceList: ["unit", "Prototype-Stubs"],
+                extraText: "some extra data",
+                messages: [],
+              };
+            }
+
+            return {
+              choiceKind: "Keyword",
+              choiceList: [], // ChoiceList for "auto vmock_session = " returns an empty list for now
+              extraText: "",
+              messages: [],
+            };
+          }
+        ),
+      };
+    });
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  test(
+    "validate coded test completion for lines of interest (// vmock) and extra text. ",
+    async () => {
+      const tstText = unitTst;
+      const lineToComplete = "// vmock";
+
+      const completionPosition = getCompletionPositionForLine(
+        lineToComplete,
+        tstText
+      );
+
+      // Flag that we are dealing with coded tests
+      const cppTestFlag = { cppNode: true, lineSoFar: lineToComplete };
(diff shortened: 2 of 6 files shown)


server/pythonUtilities.ts
```diff
@@ -1,10 +1,10 @@
 import fs = require("fs");
 import path = require("path");
-import { execSync } from "child_process";
-import { cleanVcastOutput } from "../src-common/commonUtilities";
+import { execSync } from "node:child_process";
```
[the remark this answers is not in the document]
Den1552: So I should revert all "non-xo" changes from `pythonUtilities.ts` , `setup-unit-tests.ts` and `utils.ts` ?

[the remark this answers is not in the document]
Zbigor: I guess you mean the other way around. Keep the changes you really need to get things to work, **get rid of the formatting/style changes on the source code** (server/pythonUtilities.ts)

tests/unit/tst-completion.test.ts
```diff
@@ -7,7 +7,7 @@ import {
   getCompletionPositionForLine,
   generateCompletionData,
   storeNewDocument,
-} from "./utils";
+} from "./utils.js";
```
Zbigor, line 10: why are we importing from `./utils.js` here?

Den1552: I think this happens when I execute `xo --fix`. Will change.

server/pythonUtilities.ts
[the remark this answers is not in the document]
Den1552: Reverted formatting of source code in 63d206606b08ae009c8006acf1b5dd547fea5a67

tests/unit/tst-completion.test.ts
Den1552: Deleted js from import in 4074ada0b208018d6f3865ef5c780b7e45b9ab72

aytey: If `xo --fix` is doing that, can we work out how to turn it off in the .xoconfig?

tests/unit/ct-completion.test.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+`;
+
+const timeout = 30_000; // 30 seconds
+
+describe("Testing pythonUtilities (valid)", () => {
+  beforeEach(() => {
+    // To achieve 100% coverage
+    // --> Need to mock getChoiceDataFromPython, since "extraText" is not implemented yet
+    vi.mock("../../server/pythonUtilities", async () => {
```
aytey, line 40: I don't think any other test mocks this stuff; I think we want it to run the _real_ code and get the completion stuff back.

Den1552: Added one test without mocking and extraText in 3c5d87311155ac78144ae54462bfa86fb762002d .

tests/unit/setup-unit-test.ts
```diff
@@ -126,6 +129,17 @@ module.exports = async () => {
     }
   }
 
+  const cppFilePath = path.join(vcastEnvPath, cppFilename);
+  const createCppFile = `echo -- Environment: TEST > ${cppFilePath}`;
```
aytey, line 133: The C++ files don't use this syntax

tests/unit/utils.ts
```diff
@@ -90,6 +92,7 @@ export function generateCompletionData(
   tstText: string,
   position: CompletionPosition,
   triggerCharacter: string | undefined,
+  cppTest: { cppNode: boolean; lineSoFar: string },
```
aytey, line 95: Let's make this be an interface, so we can have nullable args and not worry about always passing all of this stuff through.

tests/unit/ct-completion.test.ts
```diff
@@ -0,0 +1,145 @@
...
+      const triggerCharacter = lineToComplete.at(-1);
+
+      const codedTestCompletionData = generateCompletionData(
+        tstText,
+        completionPosition,
+        triggerCharacter,
+        cppTestFlag
+      );
+
+      expect(codedTestCompletionData).toEqual([
```
aytey, line 128: Okay, I think we need to "go deeper" on the completion.

We need to send _three_ completitons:

* `// vmock` (we should check we get a list of units)
* `// vmock unit` (we should check we get `bar`; the only function in `unit`)
* `// vmock unit bar` (we should check we get the mock snippet; see below)

For the last one, running:

```
/home/avj/vcast/vc24__101394_store_mock_info/vpython \
    /home/avj/clones/vector-vscode-vcast/coded_mock/python/testEditorInterface.py \
    choiceList-ct \
    /home/avj/clones/vector-vscode-vcast/coded_mock_tests_denis/tests/unit/vcast/TEST \
    "// vmock unit bar"
```

Gives:

```
{
  "choiceKind": "Snippet",
  "choiceList": [
    "\nint vmock_unit_bar(::vunit::CallCtx<> vunit_ctx, int z) {\n  // Enable Stub: vmock_unit_bar_enable_disable(vmock_session);\n  // Disable Stub: vmock_unit_bar_enable_disable(vmock_session, false);\n\n  // Insert mock logic here!\n}\nvoid vmock_unit_bar_enable_disable(vunit::MockSession &vmock_session, bool enable = true) {\n    using vcast_mock_rtype = int ;\n    vcast_mock_rtype (*vcast_fn_ptr)(int)  = &bar;\n    vmock_session.mock <vcast_mock_rtype (*)(int)> ((vcast_mock_rtype (*)(int))vcast_fn_ptr).assign (enable ? &vmock_unit_bar : nullptr);\n}\n// end of mock for: vmock_unit_bar -------------------------------------------------------------------------------------\n\n"
  ],
  "extraText": "",
  "messages": []
}
```

tests/unit/utils.ts
```diff
@@ -121,9 +125,24 @@ export function generateCompletionData(
...
+    if (enviroPath) {
+      return getCodedTestCompletionData(
+        cppTest.lineSoFar,
+        completion,
+        enviroPath
+      );
+    }
+    throw new ReferenceError("enviroPath is undefined.");
+
+    // TST test
```
aytey, line 141: Bad comment

tests/unit/utils.ts
```diff
@@ -121,9 +125,24 @@ export function generateCompletionData(
...
+  } else {
+    console.log(`Input .tst script: \n ${textDocument.getText()} \n`);
```
aytey, line 143: I guess this should be `.cpp file`?

tests/unit/python-utilities-valid.test.ts
```diff
@@ -68,4 +74,14 @@ describe("Testing pythonUtilities (valid)", () => {
     },
     timeout
   );
+
+  test(
```
aytey, line 78: Why is this stuff in this branch?

tests/unit/setup-unit-test.ts
Den1552: Deleted in 9038de54b41c65eec73f8f620cbae868eb5a3936 .

tests/unit/utils.ts
Den1552: Made an interface of the optional params in 9038de54b41c65eec73f8f620cbae868eb5a3936 .

tests/unit/ct-completion.test.ts
Den1552: Added test cases for the other completions in 9038de54b41c65eec73f8f620cbae868eb5a3936 .

tests/unit/utils.ts
Den1552: Changed comment in 9038de54b41c65eec73f8f620cbae868eb5a3936 .

tests/unit/utils.ts
Den1552: Switched to .cpp file in 9038de54b41c65eec73f8f620cbae868eb5a3936.

tests/unit/python-utilities-valid.test.ts
Den1552: Because pythonUtilities.ts changed in this branch (in comparison to main).
For example: 
 
https://github.com/vectorgrp/vector-vscode-vcast/blob/9861c8c9bbb6d802b6bbea5c7a981a87f091bb74/server/pythonUtilities.ts#L17-L19 
 
Therefore, I no longer had 100% coverage on this branch. Since one of my tasks is to achieve 100% coverage, I thought, 'Alright, this branch will eventually merge into main, and with that, my tests for this case will too, thereby keeping 100% coverage.'

aytey: Sorry, yes, you're correct -- perfect!

tests/unit/ct-completion.test.ts
```diff
@@ -0,0 +1,243 @@
...
+      expect(codedTestCompletionData).toEqual(extraTextMockExpected);
+    },
+    timeout
+  );
+
+  test(
+    "validate coded test completion for lines of interest (// vmock) and without extra text. ",
+    async () => {
+      // No mock for this test to let it excercise the actual python but without extra text
+      const tstText = unitTst;
```
aytey, line 164: Should we make this all be a function with args such as `lineToComplete` + `expectedReturn`? I think all of these (non-mocked) tests now have the same body, but with different inputs/expected, so I would refactor them out into a function that can be called by the tests

tests/unit/tst-completion.test.ts
```diff
@@ -1030,7 +1033,8 @@ describe("Text Completion", () => {
       const generatedCompletionData = generateCompletionData(
         tstText,
         completionPosition,
-        triggerCharacter
+        triggerCharacter,
+        cppFlagForTstTest
```
aytey, line 1037: Now it is an interface, can't we omit this parameter (apart from when we're actually doing a C++ test)?

tests/unit/ct-completion.test.ts
```diff
@@ -0,0 +1,243 @@
+import { describe, expect, test, vi, afterEach } from "vitest";
+import { getCompletionPositionForLine, generateCompletionData } from "./utils";
+
+// Whole coded test
+
+const unitTst = `
```
aytey, line 6: I think this file can likely be empty; it doesn't need to have all of this stuff in it.

If it can be empty, let's use an empty file.

tests/unit/tst-completion.test.ts
Den1552: Yeah don't know why I didn't delete it in those cases.. deleted in 805ebe742f3dfb6d82093c3d9606af7a90511b9f .

tests/unit/ct-completion.test.ts
Den1552: Used empty file in 805ebe742f3dfb6d82093c3d9606af7a90511b9f .

tests/unit/ct-completion.test.ts
Den1552: Refactored code in 805ebe742f3dfb6d82093c3d9606af7a90511b9f .

tests/unit/utils.ts
```diff
@@ -121,9 +129,25 @@ export function generateCompletionData(
...
+        optParameters.lineSoFar,
+        completion,
+        enviroPath
+      );
+    }
+
+    throw new ReferenceError("enviroPath is undefined.");
+
+    // TST
+  } else {
```
aytey, line 147: can you make this be `else /* tst */` and remove the line above?

tests/unit/setup-unit-test.ts
```diff
@@ -74,7 +75,12 @@ module.exports = async () => {
     path.join(vcastEnvPath, "unit.cpp")
   );
 
-  const clicastTemplateCommand = `cd ${vcastEnvPath} && ${clicastExecutablePath.trimEnd()} -l C template GNU_CPP11_X`;
+  const clicastTemplateCommand = `cd ${vcastEnvPath} && ${clicastExecutablePath.trimEnd()} -l C template GNU_CPP11_X option VCAST_CODED_TESTS_SUPPORT TRUE`;
+
+  await promisifiedExec(
+    `cd ${vcastEnvPath} &&  ${clicastExecutablePath.trimEnd()} -lc option VCAST_CODED_TESTS_SUPPORT TRUE`
+  );
```
aytey, line 82: We likely only want to do this when we're vc24

aytey: Also, we should make this work like the other calls (i.e., log the stderr if it exists).

tests/unit/utils.ts
```diff
@@ -90,9 +92,14 @@ export function generateCompletionData(
...
+  optParameters?: {
+    lineSoFar?: string;
+    cppTest?: boolean;
+    envName?: string;
+  }
 ) {
-  envName ||= "vcast";
+  // OptParams can be undefined
+  const envName = optParameters?.envName ?? "vcast";
   const languageId = "VectorCAST Test Script";
```
aytey, line 103: I think if this is a coded test, we should be setting the `languageId` to be `cpp` (but we should confirm actually what it should be).

tests/unit/utils.ts
Den1552: Changed comment in 770736c06264d30fd8e6d29f6695849bfad10960 .

tests/unit/setup-unit-test.ts
Den1552: Reworked in 770736c06264d30fd8e6d29f6695849bfad10960 .

tests/unit/utils.ts
Den1552: Put in cpp as languageId in case of coded tests in 770736c06264d30fd8e6d29f6695849bfad10960 .

tests/unit/setup-unit-test.ts
```diff
@@ -74,14 +76,42 @@ module.exports = async () => {
...
+    "DATA",
+    "tool_version.txt"
+  );
+  const toolVersion = fs.readFileSync(toolVersionPath).toString().trim();
+
+  // Activate coded tests support only for version 24
+  if (!toolVersion.startsWith("23")) {
+    const { stdout: stdoutClicastOption, stderr: stderrClicastOption } =
+      await promisifiedExec(
+        `cd ${vcastEnvPath} && ${clicastExecutablePath.trimEnd()} -lc option VCAST_CODED_TESTS_SUPPORT TRUE`
```
aytey, line 104: Shall we make this a string, so we can log the string later on if there's an error?

Den1552: Refactored in cb0d4bcd97809b8bcf95bf9ab4e95a4b97fba3b6.

tests/unit/ct-completion.test.ts
```diff
@@ -0,0 +1,168 @@
...
+describe("Testing pythonUtilities (valid)", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  test(
+    "validate coded test completion for lines of interest (// vmock) and extra text.",
+    async () => {
+      await validateCodedTestCompletion("// vmock", extraTextMockExpected, {
+        mockReturnValue: {
```
aytey, line 99: Let's add a comment to make it clear that this test is "different" because it mocks the Python utilities (also make it clear why we're doing that).

tests/unit/setup-unit-test.ts
```diff
@@ -75,13 +77,38 @@ module.exports = async () => {
...
+  if (stderrClicast) {
+    console.log(stderrClicast);
     throw new Error(`Error when running ${clicastTemplateCommand}`);
   }
 
-  console.log(stdout);
+  console.log(stdoutClicast);
+
+  // Construct the path to the tool_version.txt file
+  const toolVersionPath = path.join(
```
aytey, line 92: Shouldn't we be using your `getToolVersion` function here?

Den1552: Used `getToolVersion` and replaced it to `utils.ts` in 27ab614ad1b4ca582f56f93af2111503196cfd58.

tests/unit/ct-completion.test.ts
Den1552: Added comment in 27ab614ad1b4ca582f56f93af2111503196cfd58.

PR #169, opened 2024-08-08
Updated updated VectorCAST API for mocks

aytey:
Move towards the updated API in VectorCAST for coded mocks


python/dataAPIutilities.py
```diff
@@ -4,48 +4,16 @@
...
 # Tag for the init, which we want to ignore
 TAG_FOR_INIT = "<<INIT>>"
 
-# list of functions not to be shown in the functions list
-FUNCTIONS_TO_IGNORE = {"coded_tests_driver", TAG_FOR_INIT}
 
-
-def functionCanBeMocked(functionObject):
-    """
-    # PCT-FIX-NEEDED - issue #7 - is_mockable not dependable
```
johnpaliotta, line 17: Are all of these special cases supported by dataAPI now?
Are issues 7 and 8 fixed?  They are not marked as tested or fixed in confluence.

python/dataAPIutilities.py
```diff
@@ -4,48 +4,16 @@
 """
```
johnpaliotta, line 4: Agree with your suggestion to get rid of this package and move the remaining stuff to tstUtilities.py

TAG_FOR_INIT can go there too, no need for a separate file.

aytey: Doing this will create a dependency between `vTestInterface.py` and `tstUtilities.py` -- are you okay with that?

python/dataAPIutilities.py
aytey: I will check; I tested `vmock_bugs` and `vmock_examples` with the new logic and it didn't try to mock anything it shouldn't have done.

PR #173, opened 2024-08-09
Ignore environment directories at workspace root

### LINKED ISSUE #172: The extension should ignore environment directories that are opened directly.
If your workspace root is a VectorCAST environment directory, the test pane will have an environment node with no name, and some commands like Re-Build Environment will fail because the directory cannot be deleted when it is open in the Workspace!

This edge case should not have been supported in the first place.

johnpaliotta:
Also fixed an "extra" update that was being done on insert ATG tests

Closes #172

### DIFF
--- a/src/testPane.ts
+++ b/src/testPane.ts
@@ -281,6 +281,19 @@ function processVCtestData(
   }
 }
 
+function getWorkspaceFolderList(): string[] {
+  // This function will covert the workspace folders into a 
+  // string list of folders to be used in getEnvironmentList
+
+  let returnList: string[] = [];
+  for (const workspace of vscode.workspace.workspaceFolders || []) {
+    const workspaceRoot = workspace.uri.fsPath;
+    // doing the path separate replacement here to avoid compexity bwlow
+    returnList.push(workspaceRoot.replaceAll("\\", "/"));
+  }
+  return returnList;
+}
+
 const glob = require("glob");
 function getEnvironmentList(baseDirectory: string): string[] {
   // This function will find all of the VectorCAST and vTest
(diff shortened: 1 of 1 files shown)


aytey:
Can we add a notification that says something like "the current workspace is a VectorCAST environment; to open-up this environment, please open the parent folder as the workspace"?

src/testPane.ts
```diff
@@ -292,10 +305,19 @@ function getEnvironmentList(baseDirectory: string): string[] {
...
   // so turn this into a list of the enclosing directories only
   let returnList: string[] = [];
+  const workspaceFolderList = getWorkspaceFolderList();
   for (const file of fileList) {
     // note that glob always uses / as path separator ...
     if (!file.includes(".BAK")) {
-      returnList.push(path.dirname(file));
+      const candidatePath = path.dirname(file);
+      // we don't want to support users who add an enviro
+      // directory to the workspace - someone did this :()
```
aytey, line 314: Do you want a `:)` or a `:(` here? :D

src/testPane.ts
```diff
@@ -292,10 +305,19 @@ function getEnvironmentList(baseDirectory: string): string[] {
...
+      if (!workspaceFolderList?.includes(candidatePath)) {
+        returnList.push(candidatePath);
+      } else {
+        vectorMessage (`Ignoring environment: ${candidatePath} ...`);
+        vectorMessage (`   environments should not be added to the workspace.`);
```
aytey, line 319: "environments should not be opened as the workspace root"?

The current wording reads like "never open up a workspace that contains environments" (which is clearly wrong).

johnpaliotta:
> @johnpaliotta do you want a test for this?

not needed

aytey (approved):

PR #180, opened 2024-08-19
Coded mock version checks

aytey:
This PR centralises and cleans-up a lot of our versions checks, and tries to put them all in one file.

This passes the E2E tests for coded mocks.

### DIFF
--- a/python/tstUtilities.py
+++ b/python/tstUtilities.py
@@ -606,7 +606,7 @@ def getUnitAndFunctionObjects(api, unitString, functionString):
                 returnUnitList.append(unitObject)
 
     # if coded mocks are enabled, and the unit name is an exact match, process the function name
-    if enviroSupportsMocking(api) and len(returnUnitList) == 1:
+    if len(returnUnitList) == 1:
         # check if the function name matches any of the functions in the unit
         for functionObject in unitObject.functions:
             if functionCanBeMocked(functionObject):
--- a/python/tstUtilities.py
+++ b/python/tstUtilities.py
@@ -16,37 +16,17 @@ from enum import Enum
 from vector.apps.DataAPI.unit_test_api import UnitTestApi
 from vector.apps.DataAPI.unit_test_models import Function, Global
 
-CODED_MOCK_ENABLED = False
-try:
-    from vector.apps.DataAPI import mock_helper
-
-    # NOTE: we don't need to check if an environment supports coded mocks, only
-    # a version of _VectorCAST_.
-    #
-    # If an environment doesn't have coded tests enabled (or it was originally
-    # build with a version of VectorCAST that did not support coded tests/coded
-    # mocks) and we open it up with a version of VectorCAST with `mock_helper`,
-    # then `.mock.` will be set to `None`.
-    #
-    # We also need to check that the verison of the api in `mock_helper` is
-    # compatible with this version of the extension.
-
-    # The version of the api this version of the extension supports
-    supported_mock_api_version = 1
-
-    # We only enable coded mocks if the version of `mock_helper` is <= than our
-    # supported version
-    CODED_MOCK_ENABLED = mock_helper.MOCK_API_MAJOR <= supported_mock_api_version
-
-except ImportError:
-    pass
-
 from vConstants import (
     TAG_FOR_INIT,
     TAG_FOR_GLOBALS,
     CODED_TEST_SUBPROGRAM_NAME,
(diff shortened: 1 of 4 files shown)


python/tstUtilities.py
```diff
@@ -626,7 +606,7 @@ def getUnitAndFunctionObjects(api, unitString, functionString):
                 returnUnitList.append(unitObject)
 
     # if coded mocks are enabled, and the unit name is an exact match, process the function name
-    if CODED_MOCK_ENABLED and len(returnUnitList) == 1:
+    if enviroSupportsMocking(api) and len(returnUnitList) == 1:
```
johnpaliotta, line 609: Shouldn't there be a check way before this for enviroSupportsMocking?
For example, the extension should not be even calling the ct completion if the enviro does not support coded mocks
If we need this for the mock generator, then we should move it there

python/vTestInterface.py
```diff
@@ -174,11 +173,26 @@ def generateTestInfo(enviroPath, test):
...
+
+def getEnviroSupportsMock(enviroPath):
+    """
+    The extension needs to know if the environment was built with mocking
+    support, not just if the tool supports it.
+
+    If the enviro was not built with mocking then the new mocking fields in the
+    API will be set to None by the migration process.
+    """
+    try:
```
johnpaliotta, line 185: Are you assuming that the call to get the API will do the migrate?  IF so add that as a comment.

Note I have a bug right now where the data api is not migrating the cover db version

python/versionChecks.py
```diff
@@ -0,0 +1,92 @@
...
+    mocks) and we open it up with a version of VectorCAST with `mock_helper`,
+    then `.mock.` will be set to `None`.
+    """
+
+    retVal = False
+    try:
+        # Does the package exist?
+        from vector.apps.DataAPI import mock_helper
+
+        # The version of mock helper this version of the extension supports
```
johnpaliotta, line 55: Add a comment that we are hard-coding etc.
So over time we will might have to add our own translation layer here.
How can we stay in sync? 
It seems like we should have some assert that kills our testing if we ever see a version > our version

python/versionChecks.py
```diff
@@ -0,0 +1,92 @@
+import functools
+
+from vector.enums import ENVIRONMENT_VERSION_TYPE_T
```
johnpaliotta, line 3: Does this exist in older version of dataAPI, I thought Richard just added this.

python/versionChecks.py
```diff
@@ -0,0 +1,92 @@
...
+def enviroSupportsMocking(api):
+    """
+    Determine if the current _environment_ supports coded mocks.
+
+    The current environment supports mocking if vpython supports mocking, and
+    the environment is built at at least 'REVISION_2024_CODED_MOCK_DATA'
+
+    NOTE: vpythonHasCodedMockSupport() is False if ENVIRONMENT_VERSION_TYPE_T
+    does not have REVISION_2024_CODED_MOCK_DATA, so it is safe to try accessing
+    this attribue (i.e., it will not raise an attribute error)
```
johnpaliotta, line 85: because this is only called if vpythonHasCodedMockSupport is true?

python/versionChecks.py
aytey: Richard added `REVISION_2024_CODED_MOCK_DATA` to `ENVIRONMENT_VERSION_TYPE_T`, but `ENVIRONMENT_VERSION_TYPE_T` has existed for a while.

python/versionChecks.py
aytey: Because Python has short-circuiting, if it is `False` for `vpythonHasCodedMockSupport`, then it never evaluates `VIRONMENT_VERSION_TYPE_T.REVISION_2024_CODED_MOCK_DATA`, so we don't see the `AttributeError`.

python/versionChecks.py
aytey: > Add a comment that we are hard-coding etc.
> So over time we will might have to add our own translation layer here.

Done.

> How can we stay in sync?

Problem for a later date when the API actually has "breaking changes".

> It seems like we should have some assert that kills our testing if we ever see a version > our version

`vmockGenerator.py` will "immediately fail" if `vpythonHasCodedMockSupport()` returns `False` (which is this function).

python/vTestInterface.py
aytey: As discussed; this is not being addressed as how/when DataAPI/CoverDB migrates is not standardised.

PR #185, opened 2024-08-26
Clean-up unit tests for `clicast_server`

Den1552:
## Summary 
Fix unit tests for `clicast_server` and added tests to get coverage back to 100%. 

## Changed
- Updated several tests to achieve 100% coverage.

## Added
- New file `vite.config.utils.ts` for  `getToolVersion() `, since trying to 
   share it into `vite.config.ts` causes problems with `node-fetch` 
   import.
- ` vcast-server.test.ts`, testing methods from `vcast-server.ts`


tests/unit/python-utilities-invalid.test.ts
```diff
@@ -48,16 +49,30 @@ describe("Testing pythonUtilities (invalid)", () => {
         "/some/command",
       ];
 
-      // Mock existsSync since path does not exist
-      existsSyncSpy = vi
-        .spyOn(fs, "existsSync")
-        .mockImplementation(() => false);
+      // Call the function
```
johnpaliotta, line 52: Can we never insert a comment like this please - it tells us nothing

src-common/vcastServer.ts
```diff
@@ -1,6 +1,9 @@
 // proof of concept for interacting with a running environment data server
 
-import fetch from "node-fetch";
+const fetch = (...args: [string, RequestInit?]) =>
```
johnpaliotta, line 3: What is this and why is it changing in a PR to fix compile errors in UTs

tests/unit/python-utilities-invalid.test.ts
```diff
@@ -9,11 +9,12 @@ import {
   afterEach,
   type SpyInstance,
 } from "vitest";
+import { getChoiceData } from "../../server/pythonUtilities";
 
 const timeout = 30_000; // 30 seconds
 
 let consoleLogSpy: SpyInstance;
-let existsSyncSpy: SpyInstance;
+// let existsSyncSpy: SpyInstance;
```
johnpaliotta, line 17: Don't comment out code and push it.
If there is a valid reason to comment out something to save it, add an explicit comment

tests/unit/python-utilities-valid.test.ts
```diff
@@ -48,20 +46,20 @@ describe("Testing pythonUtilities (valid)", () => {
...
+        "..",
+        "..",
+        "python",
+        "testEditorInterface.py"
+      );
 
-      const validPathToTestEditorInterface =
-        "some/valid/path/python/testEditorInterface.py";
-      const expectedMessagePart = `testEditorInterface was found here: ${validPathToTestEditorInterface}`;
+      // Call the function
```
johnpaliotta, line 59: Same as above

src-common/vcastServer.ts
aytey:
```
node_modules/vite/node_modules/esbuild/lib/main.js:1373:27: ERROR: [plugin: externalize-deps] "node-fetch" resolved to an ESM file. ESM file cannot be loaded by `require`. See http://vitejs.dev/guide/troubleshooting.html#this-package-is-esm-only for more details.
```

https://stackoverflow.com/a/69331469

aytey: https://github.com/node-fetch/node-fetch/blob/HEAD/docs/v3-UPGRADE-GUIDE.md

```
const fetch = (...args) => import('node-fetch').then(({default: fetch}) => fetch(...args));
```

aytey: To be clear: I don't know the difference between how the tests run and how the extension is built, and why this seems okay when building the extension.

aytey: Okay, got it. The problem is that we're trying to share `getToolVersion` into `vite.config.ts`, which causes this issue.

@Den1552: let's put `getToolVersion` in its own file and minimise what gets imported into that new file. Hopefully that solves this problem.

Den1552: Refactored in 8fdcb1cdf2798b7189e78e16f572f035c8fa8f64 .

tests/unit/python-utilities-valid.test.ts
Den1552: Removed comment in 8fdcb1cdf2798b7189e78e16f572f035c8fa8f64 .

tests/unit/python-utilities-invalid.test.ts
Den1552: Removed commented code in 8fdcb1cdf2798b7189e78e16f572f035c8fa8f64 .

tests/unit/python-utilities-invalid.test.ts
Den1552: Removed comment in 8fdcb1cdf2798b7189e78e16f572f035c8fa8f64 .

tests/unit/setup-unit-test.ts
```diff
@@ -3,7 +3,7 @@ import process from "node:process";
 import { exec } from "node:child_process";
 import { promisify } from "node:util";
 import { rm, mkdir, copyFile } from "node:fs/promises";
-import { getToolVersion } from "./utils";
+import { getToolVersion } from "../../vite.config.utils";
```
aytey, line 6: Is this the base name for it? Why not `tests/getToolVersion.ts` or something?

tests/unit/setupUnitTest.ts
```diff
@@ -110,7 +110,8 @@ module.exports = async () => {
   const tstFilePath = path.join(vcastEnvPath, tstFilename);
   const createTstFile = `echo -- Environment: TEST > ${tstFilePath}`;
   {
-    const stderr = await promisifiedExec(createTstFile)[1];
+    const result = await promisifiedExec(createTstFile);
```
aytey, line 113: Why not:

```
const { stdout, stderr } = await promisifiedExec(checkClicast);
```

?

tests/unit/setup-unit-test.ts
Den1552: Renamed in e692cf4675eb7ccf870c3d73c1df16f09988653c .

tests/unit/setupUnitTest.ts
Den1552: Refactored code in e692cf4675eb7ccf870c3d73c1df16f09988653c. Added `runCommand` function in `utils` to reduce code duplication.

tests/unit/ct-completion.test.ts
```diff
@@ -134,6 +225,216 @@ describe("Testing pythonUtilities (valid)", () => {
...
+    "validate console log if connection is defined",
+    async () => {
+      // Create a custom connection object with a console.log function
+      const customConnection = {
+        console: {
+          log: (message: string) => console.log(message),
+        },
+      };
+
+      const lineToComplete = "// vmock";
```
Zbigor, line 239: 239-280 could be a function like `prepareCodedTestCompletion` (312-346 is almost exactly the same, so it might be worth extracting the preparation logic to a separate function if we're going to have similar tests in the future)

tests/unit/python-utilities-valid.test.ts
```diff
@@ -80,4 +44,40 @@ describe("Testing pythonUtilities (valid)", () => {
     },
     timeout
   );
+
+  test("should create and send a diagnostic objects for tst", () => {
+    const diagnostic: Diagnostic = {
```
Zbigor, line 49: Have a look if you can make a utility function for diagnostics stuff - quite some things seem to be repeated from `ct-completion.test.ts`

Zbigor (changes requested):
A couple of minor refactoring opportunities in there for you to have a look at but looks good overall

tests/unit/ct-completion.test.ts
Den1552: Refactored in 314cad1a860f1d3471af84c9d50e0bfc08afd811 .

tests/unit/python-utilities-valid.test.ts
Den1552: Added setupDiagnosticTest in 314cad1a860f1d3471af84c9d50e0bfc08afd811 .

Zbigor (approved):

PR #187, opened 2024-08-29
Add support for `clicast` server mode

johnpaliotta:
VectorCAST 2024sp4 has support for running `clicast` in a server mode.

This PR adds support to the extension to utilise this server. Currently, the server is not enabled by default.


python/pythonUtilities.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+    # A valid clicast server start emits: "clicast-server-started"
+    # so we check for that, or we wait for the process to terminate
+    # which a non-server clicast version will do.
+    # Note that this is a failsafe because we should never get
+    # here if the clicast is not server capable.
+    while True:
+        responseLine = processObject.stdout.readline()
+        if responseLine.startswith("clicast-server-started"):
```
Zbigor, line 63: Do we want to add a timeout for this check, or we're sure that the process will terminate in a reasonable time anyway?

python/pythonUtilities.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def cleanEnviroPath(enviroPath):
+    """
+    This function is used to clean up the environment path to make it usable as
+    a consistent dictionary key.  We force the drive lever to lower case
+    and we replace backslashes with forward slashes.
+    """
+    returnPath = enviroPath.replace("\\", "/")
+    if returnPath[2] == ":":
```
Zbigor, line 155: I guess we'd want to check `returnPath[1]` for stuff like `C:/path/to/some/env`

python/vTestInterface.py
```diff
@@ -511,15 +520,18 @@ def validateClicastCommand(command, mode):
...
         ):
             pass
         else:
-            print(f"Invalid value for --clicast: {command}")
-            raise UsageError()
+            raise UsageError("--clicast argument is invalid, file does not exist")
+
+
+def validatePath(pathString):
+    if not os.path.isdir(pathString) and not os.path.isfile(pathString):
```
Zbigor, line 533: I guess we want `or` instead of `and` here?

src-common/vcastServer.ts
```diff
@@ -0,0 +1,221 @@
...
+let HOST = "localhost"; // The server's hostname or IP address
+let PORT = 60461; // The port used by the server anything > 1023 is OK
+
+// Types used for interaction with the python interface and the enviro server
+//
+// Note: some of these match the --mode argument for the vTestInterface.py but
+// some of these (like ping) are for the enviro server only
+//
+// Note: this enum must stay in sync with the server file: vcastDataServerTypes.py: commandType.
+
```
Zbigor, line 13: Do we think this enum will be changing frequently? Maybe we could have a `.json` file with this mapping that gets read by both `.ts` and the `.py` counterpart. If the enum doesn't really change often, it might not be worth moving the mapping to a json file

src/client.ts
```diff
@@ -160,6 +170,18 @@ export function sendVPythonCommandToServer(vPythonCommand: string) {
   }
 }
 
+// This function is used to send an updated state of the
+// enviro data server to the language server
+export function sendServerStateToServer(useServer: boolean) {
```
Zbigor, line 175: Maybe change `sendServerStateToServer` -> `sendServerStateToLanguageServer` for clarity

src/vcastAdapter.ts
```diff
@@ -88,55 +190,80 @@ export function deleteEnvironment(enviroPath: string, enviroNodeID: string) {
...
-    path.dirname(scriptPath)
-  );
+  let loadScriptArgs: string = `-e${enviroName} test script run ${scriptPath}`;
+
+  let commandStatus: commandStatusType;
+  // using server ....
+  if (globalEnviroDataServerActive) {
+    const enviroPath = path.join(path.dirname(scriptPath), enviroName);
+    commandStatus = await executeClicastCommandUsingServer(
+      clicastCommandToUse,
```
Zbigor, line 206: clicastCommandToUse argument is never used in the implementation of `executeClicastCommandUsingServer`

Zbigor (approved):
There are only a couple of minor points but the changes look good overall. We still need to have a look at those E2E tests that failed to see if those are "real bugs" or just some tests that need to be updated

python/clicastInterface.py
```diff
@@ -104,24 +172,95 @@ def runClicastScriptCommandLine(commandFileName, echoToStdout):
...
+
+    # noServer allows the caller to specify that we should run clicast directly
+
+    if pythonUtilities.USE_SERVER:
+        return runClicastScriptUsingServer(enviroPath, commandFileName)
+    else:
+        return runClicastScriptCommandLine(commandFileName, echoToStdout)
+
+
+tempEnviroScript = "rebuild.env"
```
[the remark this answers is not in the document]
johnpaliotta: Fixed with d59bb01

src/client.ts
johnpaliotta: Fixed with 07ef593

src/vcastAdapter.ts
johnpaliotta: Fixed with bdfc7e9

src-common/vcastServer.ts
johnpaliotta: I don't think we will be changing this very often.  I added a comment to both places with this commit: 1615e28

python/pythonUtilities.py
johnpaliotta: @Zbigor Great catch on this one, thanks!
Fixed with: 817e1dc

python/pythonUtilities.py
johnpaliotta: Good idea, the while true did make me a bit uncomfortable ...  improved with c4d4050

python/vTestInterface.py
johnpaliotta: I think it is ok as is because sometimes it is called with a file path and sometimes with a directory.
Let me know if I am missing something :)

Zbigor: The condition might be clearer as `not (os.path.isdir(pathString) or os.path.isfile(pathString))` but it's just personal preference really, it's correct as is

src-common/vcastServer.ts
Zbigor: All good then

PR #189, opened 2024-08-30
Added tests for clicast_server coverage

Den1552:
## Summary

Added tests to get back to 100% coverage.

## Changed
- Added exports to functions in `vcast-server-test.ts `to enable testing
- Enhanced test for `transmitCommand()`


tests/unit/vcast-server.test.ts
```diff
@@ -95,7 +98,7 @@ describe("test server functions", () => {
     );
   });
 
-  it("serverIsAlive handles Python interface error", async () => {
+  test("serverIsAlive handles Python interface error", async () => {
```
Zbigor, line 101: Can we revert to `it` for consistency with other tests? If we want to switch from `test` to `it`, it should be a separate PR that changes this in all tests inside `unit`

Zbigor (changes requested):
Please revert `test` to `it` for consistency

Zbigor (approved):

src-common/vcastServer.ts
```diff
@@ -34,7 +34,7 @@ export interface clientRequestType {
   options?: string;
 }
 
-function serverURL() {
+export function serverURL() {
```
johnpaliotta, line 37: Denis could you please add comments for these functions that are newly exported to indicate that they are being exported for testing purposes please.

Ping me after this and we can merge this.

PR #190, opened 2024-09-02
Added unit tests debugger in launch.json.example

Den1552:
## Summary

Enhanced `launch.json.example`  by adding a new debugger configuration for VectorCAST unit tests.

### DIFF
--- a/.vscode/launch.json.example
+++ b/.vscode/launch.json.example
@@ -1,7 +1,7 @@
 {
   "version": "0.2.0",
   "configurations": [
-     {
+    {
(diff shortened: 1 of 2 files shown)


Zbigor (changes requested):
With the config as is, the test runner is running all the previous tests until reaching the one that we want to debug (I tested this by setting a breakpoint on line 15 in `common-utilities.test.ts`). The goal of the config is to be able to debug a unit tests directly, without waiting for other tests to run

Zbigor (approved):

PR #193, opened 2024-09-05
Update ct support for tst #170

### LINKED ISSUE #170: Test script editing needs some updates to support coded tests
There are a few improvements needed - in the main branch

- [x] `TEST.CODED_TEST_FILE` should be added as a valid syntax in a .tst file (it should appear in the auto-complete list when you type in `TEST.`, and not give an error when you type in)
   - What comes after the `:` will be a file path (this could be relative or absolute), but we do not want to do auto-completion on the filename (i.e., for `TEST.CODED_TEST_FILE` there is no autocomplete after `:`)
   - We probably need to make sure that coded tests are enabled in the environment/version of VectorCAST as well (they are not supported in 2023)
- [x] When the `TEST.SUBPROGRAM` is set to: `coded_tests_driver`
  - [x] `TEST.CODED_TEST_FILE` should be allowed (that is, it should not give error diagnostics)
  - [x] `TEST.VALUE` and `TEST.EXPECTED` lines should not be allowed (these should give error diagnostics)
  - [x] Look for "Command is only valid" inside of the Python to see what kind of errors we already return
- [x] When the `TEST.SUBPROGRAM` is **not** set to: `coded_tests_driver`
  - [x] `TEST.CODED_TEST_FILE` should **not** be allowed (i.e., it should now give an error diagnostic)
- [x] `coded_tests_driver` should not show up in the subprogram list for `TEST.VALUE` and `TEST.EXPECTED` lines
    - Normally when you type in `TEST.VALUE:<unit>.` you will get a list of all of the subprograms in `<unit>` -- DataAPI _will tell you_ that `coded_tests_driver` is in `<unit>`, but we want to hide that from a user (the same for `TEST.EXPECTED)

## Note:

- In file `tstCompletions.ts, comment:  // TBD will need to change how this is done during the fix for issue #170`
  - The way I get the completion for `TEST.SUBPROGRAM` is a little bit of a hack and will need to change
  - Currently, I send a fake `TEST.VALUE:unit.` line to get the list of subprograms
  - But currently, the Python strips coded_tests_driver in function: `getFunctionList()`
- Need to check if coded_tests_driver should be allowed in `TEST.SLOT` lines
  - Does vcast support compounds with coded tests?
  - @Den1552 won't address this until Andrew is back
- Look for all other uses of `getFunctionList()`
  - `getFunctionList()` is the "core" of the interaction between VS Code and VectorCAST, and is probably the place where `coded_test_driver` should be filtered, but _not_ for `TEST.SUBPROGRAM`

Den1552:
## Summary

Updates script editing to support coded tests and resolves #170. 

## Added
- **Syntax Support for `TEST.CODED_TEST_FILE`:**
  - Added `TEST.CODED_TEST_FILE` as a valid syntax in `.tst` files.
  - Enabled auto-complete for `TEST.CODED_TEST_FILE` after typing `TEST.`.
  - Ensured no auto-completion for the file path after `:`.

- **Behavior When `TEST.SUBPROGRAM` is `coded_tests_driver`:**
  - Allowed `TEST.CODED_TEST_FILE` without error diagnostics.
  - Disallowed `TEST.VALUE` and `TEST.EXPECTED` lines, ensuring they give error diagnostics.

- **Behavior When `TEST.SUBPROGRAM` is Not `coded_tests_driver`:**
  - Disallowed `TEST.CODED_TEST_FILE`, ensuring it gives an error diagnostic.
  - Removed `coded_tests_driver` from the subprogram list for `TEST.VALUE` and `TEST.EXPECTED` lines.

- **Error Handling Enhancements:**
  - Reviewed and ensured consistency in error messages related to these changes.

- **Tests**
   - Added tests for new diagnostics


Den1552:
Currently working on Note: 

> In file tstCompletions.ts, comment:  // TBD will need to change how this is done during the fix for issue #170
> The way I get the completion for TEST.SUBPROGRAM is a little bit of a hack and will need to change
> Currently, I send a fake TEST.VALUE:unit. line to get the list of subprograms
> But currently, the Python strips coded_tests_driver in function: getFunctionList()

-- > Check how to get autocompletion for TEST.SUBPROGRAM without 
       hardcoding it & get correct autocompletion items.

server/tstCompletion.ts
```diff
@@ -17,18 +17,84 @@ import {
...
 
+/**
+ * Checks if a specific VectorCAST option is set to a given value in the specified environment.
+ *
+ * @param {string} enviroPath - The file path to the environment where the command should be executed.
+ * @param {string} option - The VectorCAST option to check (e.g., 'VCAST_CODED_TESTS_SUPPORT').
+ * @param {string} optionValue - The expected value of the option (e.g., 'True', ...).
+ * @returns {Promise<boolean>} - Resolves to `true` if the option value matches, otherwise `false`.
+ */
+export async function checkClicastOption(
```
Zbigor, line 35: A couple of things to consider here:
- https://github.com/vectorgrp/vector-vscode-vcast/blob/2cb2dc5ccedaa9e507174b2d2e47befce2e7788a/src/vcastInstallation.ts#L304 is definining the `clicastCommandToUse`
- https://github.com/vectorgrp/vector-vscode-vcast/blob/2cb2dc5ccedaa9e507174b2d2e47befce2e7788a/src/vcastAdapter.ts#L244 is an example 
 - you should try to re-use existing functions/keep the consistency in the repo
 - In the code above you can see ` executeCommandSync` being used for a very similar command (server variant is https://github.com/vectorgrp/vector-vscode-vcast/blob/2cb2dc5ccedaa9e507174b2d2e47befce2e7788a/src/vcastAdapter.ts#L219C27-L219C33)
 - If we are calling this for every completion, some caching might make sense
   - Something like https://www.npmjs.com/package/@type-cacheable/core with lru cache

Zbigor:
> Currently working on Note:
> 
> > In file tstCompletions.ts, comment:  // TBD will need to change how this is done during the fix for issue #170
> > The way I get the completion for TEST.SUBPROGRAM is a little bit of a hack and will need to change
> > Currently, I send a fake TEST.VALUE:unit. line to get the list of subprograms
> > But currently, the Python strips coded_tests_driver in function: getFunctionList()
> 
> -- > Check how to get autocompletion for TEST.SUBPROGRAM without hardcoding it & get correct autocompletion items.
> 
> Problem is: I need the `unit`. So either:
> 
> **Some ideas / brainstorming:**
> 
> **1)** Retrieve unit with `getNearest `in ts and add it as an additional param --> Additional param --> Very specific (what if we need in future more than just the unit in python?) --> Send JSON?
> 
> **2)** Save tst to temp file and send abs path as additional param and retrieve the unit in python --> Would be on an "logical level" the correct place imo --> Good if we need the whole tst for later stuff
> 
> **3)** Get Unit with `getNearest `in ts, `concat `it as `TEST.SUBPROGRAM:<unit>` and send it to get autocompletion --> No big implementation changes since `getFunctionList` gets called here --> Just add `<<INIT>>` and `<<COMPOUND>>` at the end --> But is weird, since `TEST.SUBPROGRAM:<unit>` is wrong and maintainability would suffer since you do not expect that it works that way.
> 
> --> Will do **1)** first, as I think that additional params can just be added in the json if needed.

Can you elaborate on the 2? Where and what would actually be implemented (python/ts, which function)? Could you pass around some string instead of having a tempfile? Whichever option you implement should be in a separate branch off of this branch

Den1552:
Can you elaborate on the 2? Where and what would actually be implemented (python/ts, which function)? Could you pass around some string instead of having a tempfile? Whichever option you implement should be in a separate branch off of this branch

> > Currently working on Note:
> > > In file tstCompletions.ts, comment:  // TBD will need to change how this is done during the fix for issue #170
> > > The way I get the completion for TEST.SUBPROGRAM is a little bit of a hack and will need to change
> > > Currently, I send a fake TEST.VALUE:unit. line to get the list of subprograms
> > > But currently, the Python strips coded_tests_driver in function: getFunctionList()
> > 
> > 
> > -- > Check how to get autocompletion for TEST.SUBPROGRAM without hardcoding it & get correct autocompletion items.
> > Problem is: I need the `unit`. So either:
> > **Some ideas / brainstorming:**
> > **1)** Retrieve unit with `getNearest `in ts and add it as an additional param --> Additional param --> Very specific (what if we need in future more than just the unit in python?) --> Send JSON?
> > **2)** Save tst to temp file and send abs path as additional param and retrieve the unit in python --> Would be on an "logical level" the correct place imo --> Good if we need the whole tst for later stuff
> > **3)** Get Unit with `getNearest `in ts, `concat `it as `TEST.SUBPROGRAM:<unit>` and send it to get autocompletion --> No big implementation changes since `getFunctionList` gets called here --> Just add `<<INIT>>` and `<<COMPOUND>>` at the end --> But is weird, since `TEST.SUBPROGRAM:<unit>` is wrong and maintainability would suffer since you do not expect that it works that way.
> > --> Will do **1)** first, as I think that additional params can just be added in the json if needed.
> 
> Can you elaborate on the 2? Where and what would actually be implemented (python/ts, which function)? Could you pass around some string instead of having a tempfile? Whichever option you implement should be in a separate branch off of this branch

Here the idea was to basically have the whole tst document in python ready. And the easiest way to have the whole document ready would be probably to save it in a temp file and then load it again in python.

Your idea of parsing a string is basically idea **1)**. Just that I do not want to parse the whole tst string as it will get executed in a command line. 

https://github.com/vectorgrp/vector-vscode-vcast/blob/e643e928f318589fa5b34902ac4c544d04f4cfe3/server/pythonUtilities.ts#L67

If the file becomes too large, it could lead to issues (I think). 

The idea is to create an object (like JSON) that carries the necessary parameters. For example, currently, I would have `{params: {unit: "moo"}}`. If more parameters are needed later, they can be easily added.

PR #195, opened 2024-09-06
New implementation for TEST.SUBPROGRAM autocompletion

Den1552:
## Summary:

This implementation should provide an autocompletion list for `TEST.SUBPROGRAM` without using the fake `TEST.VALUE` line described in #170.

## Explanation:

- In order to get the autocompletion list for `TEST.SUBPROGRAM`, we need the `<unit>` from `TEST.UNIT`.
- This is retrieved with `getNearest()` in `tstCompletion.ts`.
- This `<unit>` is sent as an additional parameter to the Python logic.
- In `processSubprogramLines()` (tstUtilities.py), we use this `<unit>` to retrieve, with `getFunctions()`, the `autocompletion` list of all the functions in this unit from the `DataAPI`.
- At the end, we add SUBPROGRAM-specific autocompletions like `<<INIT>>, <<COMPOUND>>, <<coded_tests_driver>>`.

Basically, the implementation works. However, we need to answer some questions:

## Questions:

1) Passing this argument into `getChoiceDataFromPython` is straightforward, but how about the "server-side" autocompletion retrieval in `getChoiceDataFromServer()`?  
https://github.com/vectorgrp/vector-vscode-vcast/blob/2992c9ea72a87dcbbf7e3626ea6fcb6e7b5a26b5/server/pythonUtilities.ts#L101

2) What about the SUBPROGRAM-specific autocompletions (`<<INIT>>, <<COMPOUND>>, coded_tests_driver`)?  
Is it all right to add them like that in `processSubprogramLines()`, or should this also be retrieved from the DataAPI somehow?


python/testEditorInterface.py
```diff
@@ -33,7 +33,7 @@ def main():
     # We get here when the user types a "." or ":"
 
     # argv has the name of the script as arg 1 and then user args
-    if len(sys.argv) == 4:
+    if ((len(sys.argv) == 4) or (len(sys.argv) == 5)):
```
johnpaliotta, line 36: @aytey 
These are very quick comments, as it is the end of the day and I don't have time to look at this thoroughly

This is not good, the check for 4 was to validate that the number of args was correct for the call, just slapping 5 here makes no sense, and also invalidates the else below which says 3 args are expected.

python/testEditorInterface.py
```diff
@@ -43,6 +43,12 @@ def main():
         # Contents of the line from the editor so far
         inputLine = sys.argv[3]
 
+        additionalParams = None
+
+        if(len(sys.argv) == 5):
+            # Additional autocompletion params
```
johnpaliotta, line 49: Do not write comments like this, what is saved in arg5?

python/testEditorInterface.py
```diff
@@ -52,7 +58,7 @@ def main():
                 choiceData = choiceDataType()
 
         elif mode == "choiceList-tst":
-            choiceData = processTstLine(enviroName, inputLine)
+            choiceData = processTstLine(enviroName, inputLine, additionalParams)
```
johnpaliotta, line 61: I don't even understand what is being done here yet but adding a new parameter cannot be the best way to do this

python/tstUtilities.py
```diff
@@ -417,6 +417,29 @@ def processRequirementLines(api, pieces, triggerCharacter):
 
     return returnData
 
+def processSubprogramLines(api, pieces, triggerCharacter, additionalParams):
```
johnpaliotta, line 420: Ok so if you are only using this to pass the unit name, then change all of the parameter names etc to unitName, not additionalParams.   Additionally additionalParams is a bad name because there is only one right?

python/testEditorInterface.py
```diff
@@ -43,6 +43,12 @@ def main():
         # Contents of the line from the editor so far
```
johnpaliotta, line 43: You probably did not want to tear this completely apart but I would probably convert this to use argparse now and have named parameters like --mode, --enviroName, --inputLine, --unit etc. and make unit optional, the others required.

(if it turns out we need unitName ... which it might)

server/tstCompletion.ts
```diff
@@ -38,18 +38,18 @@ export async function checkClicastOption(
...
 ): Promise<boolean> {
   const execAsync = promisify(exec);
-  const getCodedTestsSupportCommand = `${process.env.VECTORCAST_DIR}/clicast option ${option}`;
+  const getCodedTestsSupportCommand = `${process.env.VECTORCAST_DIR}/clicast get_option ${option}`;
 
-  try {
-    const { stdout } = await execAsync(getCodedTestsSupportCommand, {
-      cwd: enviroPath,
-    });
+  const { stdout, stderr } = await execAsync(getCodedTestsSupportCommand, {
```
johnpaliotta, line 43: Add a (meaningful) comment

server/tstCompletion.ts
```diff
@@ -165,24 +168,38 @@ export async function getTstCompletionData(
...
+      let choiceArray: string[] = [];
       if (unitName.length > 0) {
         const choiceData = await getChoiceData(
           choiceKindType.choiceListTST,
           enviroPath,
-          "TEST.VALUE:" + unitName + "."
+          upperCaseLine,
+          autocompletionParams
         );
+        console.log("++++++++++++++++++++++++++++++");
```
johnpaliotta, line 187: Remove

server/tstCompletion.ts
```diff
@@ -165,24 +168,38 @@ export async function getTstCompletionData(
...
+        //   choiceKindType.choiceListTST,
+        //   enviroPath,
+        //   "TEST.VALUE:" + unitName + "."
+        // );
         returnData.extraText = choiceData.extraText;
         returnData.messages = choiceData.messages;
         choiceKind = choiceData.choiceKind;
         // append actual choices to the default INIT and COMPOUND
-        choiceArray = choiceArray.concat(choiceData.choiceList);
+        // choiceArray = choiceArray.concat(choiceData.choiceList);
```
johnpaliotta, line 199: Why iis this commented out?

server/tstCompletion.ts
```diff
@@ -165,24 +168,38 @@ export async function getTstCompletionData(
...
         // <<GLOBAL>> is valid on VALUE lines but not as a function name!
-        choiceArray = filterArray(choiceArray, "<<GLOBAL>>");
+        // choiceArray = filterArray(choiceArray, "<<GLOBAL>>");
```
johnpaliotta, line 201: Same

johnpaliotta (commented):
Some very quick comments ...
Seems like some improvements could be made!

Thanks.

python/tstUtilities.py
Den1552: At the moment, we would only need the `<unit>`, but I thought to make it as an object so that for future implementations I can simply add new params to this JSON. 

I changed it now to be only `unit`. In case we will need more, I could enhance it.

python/testEditorInterface.py
Den1552: Refactored to argument parser in 78bf51f72c4d2f1a11d3f97b4d1575e0ba69e717 .

server/tstCompletion.ts
Den1552: I am sorry it was only a Draft so I didn't delete pieces of the "old" code in case you do not like my idea how I solved it. Removed in 78bf51f72c4d2f1a11d3f97b4d1575e0ba69e717 .

server/tstCompletion.ts
Den1552: Removed in 78bf51f72c4d2f1a11d3f97b4d1575e0ba69e717 .

server/tstCompletion.ts
Den1552: Removed in 78bf51f72c4d2f1a11d3f97b4d1575e0ba69e717.

python/testEditorInterface.py
Den1552: I thought this would be the best way since 
 - I need the `unit` to get all the functions for the TEST.SUBPROGRAM with `getFunctions()`.
 - The changes are not really huge 
 
 I brainstormed some ideas in the notes here: 
 https://github.com/vectorgrp/vector-vscode-vcast/pull/193
 
 I basically implemented idea **1)** because I thought it s the best of those. If you want me to, I could take again a look at it.

PR #196, opened 2024-09-11
Add vsce to package.json

aytey:
It seems we never had `vsce` in our `package.json` -- this meant that `npm run package` would likely use the global one.

This fixes that and gives us a `vsce` locally.


aytey:
@SalvatoreZagaria can you please check all our images and make sure we _do not_ have `vsce` installed globally?

PR #198, opened 2024-09-18
Fixed e2e tests for clicast_server.

Den1552:
## Summary

This PR should fix the e2e tests for clicast_server (server mode off). Most test problems were caused by: 

1) Deletion of the function `logTestResults` in clicast_server which caused many expected logs to not come up  
https://github.com/vectorgrp/vector-vscode-vcast/blob/e643e928f318589fa5b34902ac4c544d04f4cfe3/src/vcastTestInterface.ts#L416-L441

2) `errorLevel.trace` got included in some `vectorMessage()` calls, which also made some expected logs not come up 

## Modified

- Several E2E tests and mostly their loggings.
- Minor changes in other locations because of `xo`

## Added

- Configured `wdio.conf.ts` to include a `settings.json` for the tests, with `verboseLogging` set to true. This allows us to view the expected logs during the tests, marked with `errorLevel.trace` without the need to activate `verboseLogging` in every test --> saves execution time. 
- Implemented a 3-second wait after certain tests to prevent VSCode from closing too quickly, which can cause the server to terminate due to a lack of active communication.

### DIFF
--- a/tests/internal/e2e/test/specs/vcast.rest_3.test.ts
+++ b/tests/internal/e2e/test/specs/vcast.rest_3.test.ts
@@ -21,6 +21,7 @@ import {
   cleanup,
   assertTestsDeleted,
 } from "../test_utils/vcast_utils";
+import { TIMEOUT } from "../test_utils/vcast_utils";
 
 const promisifiedExec = promisify(exec);
 describe("vTypeCheck VS Code Extension", () => {
@@ -28,7 +29,6 @@ describe("vTypeCheck VS Code Extension", () => {
   let workbench: Workbench;
   let editorView: EditorView;
   let statusBar: StatusBar;
-  const TIMEOUT = 20_000;
   before(async () => {
     workbench = await browser.getWorkbench();
     // Opening bottom bar and problems view before running any tests
--- a/tests/internal/e2e/test/specs/vcast.rest_3.test.ts
+++ b/tests/internal/e2e/test/specs/vcast.rest_3.test.ts
@@ -197,10 +197,9 @@ describe("vTypeCheck VS Code Extension", () => {
     await browser.waitUntil(
       async () =>
         (await outputView.getText())
-          .at(-1)
           .toString()
-          .includes("Processing environment data for:"),
-      { timeout: 30_000, interval: 1000 }
+          .includes("Processing environment data"),
+      { timeout: TIMEOUT }
     );
     await browser.pause(10_000);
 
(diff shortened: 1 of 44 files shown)


tests/internal/e2e/test/specs/vcast.rest_3.test.ts
```diff
@@ -198,10 +197,9 @@ describe("vTypeCheck VS Code Extension", () => {
     await browser.waitUntil(
       async () =>
         (await outputView.getText())
-          .at(-1)
           .toString()
-          .includes("Processing environment data for:"),
-      { timeout: 30_000, interval: 1000 }
```
Zbigor, line 204: was the change in the way we poll necessary? we used to check every 1000 ms. If you need a longer timeout, you could just increase it. On the other hand, since the e2e is now passing, it doesn't really matter

Zbigor (approved):
all e2e passing, looks good

tests/internal/e2e/test/specs/vcast_coded_tests.test.ts
```diff
@@ -27,7 +27,7 @@ describe("vTypeCheck VS Code Extension", () => {
   let bottomBar: BottomBarPanel;
   let workbench: Workbench;
   let statusBar: StatusBar;
-  const TIMEOUT = 120_000;
```
johnpaliotta, line 30: Add a comment please about why we need this.
I realize you only bumped it from 2 to 3 but check with Igor about why its there anyway

tests/internal/e2e/test/wdio.conf.ts
```diff
@@ -242,7 +242,7 @@ export const config: Options.Testrunner = {
   //
   // Default timeout in milliseconds for request
   // if browser driver or grid doesn"t send response
-  connectionRetryTimeout: 22_000,
+  connectionRetryTimeout: 40_000,
```
johnpaliotta, line 245: Maybe create constant variables for all timeouts and put them together with unique names and descriptions and then use these vs hard-code

tests/internal/e2e/test/specs/vcast_coded_tests.test.ts
Den1552: This timeout increase is primarily for my setup, as my VM is somewhat slow during E2E execution. Otherwise, the tests fail locally.

tests/internal/e2e/test/wdio.conf.ts
Den1552: Changed `TIMEOUT` to be imported from `vcast_utils.ts` in 27896d0687ae69b062982c784885ea857156c279.

PR #199, opened 2024-09-18
Changed expected error code for compileError.

Den1552:
## Summary

This PR is related to #198. It contains a small change in the source code to fix issues when retrieving compile errors.

### DIFF
--- a/src/vcastTestInterface.ts
+++ b/src/vcastTestInterface.ts
@@ -446,7 +446,7 @@ export async function runVCTest(enviroPath: string, nodeID: string) {
   let commandOutputText: string = commandStatus.stdout;
   let executionDetails: executeOutputType = nullExecutionStatus;
 
-  if (commandStatus.errorCode == pythonErrorCodes.codedTestCompileError) {
+  if (commandStatus.errorCode == 228) {
     const testNode = getTestNode(nodeID);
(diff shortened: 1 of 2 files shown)


Den1552:
## Explanation

The issue occurs in this line:

https://github.com/vectorgrp/vector-vscode-vcast/blob/c4d4050af100e4571346010cd44d9b058ed9727a/src/vcastTestInterface.ts#L449

When a `compileError `is encountered, we expect an error message to appear for the user and the `ACOMPILE.LIS` to open. However, neither of these actions is happening, which is why the spec group `coded_tests` is failing.

The root cause is that when a compile error occurs, the expected error code (`pythonErrorCodes.codedTestCompileError (996)`) is not being returned. Instead, the error code returned is `228`. As a result, the `returnStatus `is incorrectly set to `didNotRun`.

https://github.com/vectorgrp/vector-vscode-vcast/blob/c4d4050af100e4571346010cd44d9b058ed9727a/src/vcastTestInterface.ts#L465C5-L465C41

This causes the `TextMessage` to not be displayed:

https://github.com/vectorgrp/vector-vscode-vcast/blob/75b08eee8fe72530a5a295fbea8413bbd6c9d181/src/testPane.ts#L853-L856

Ultimately, this leads to the failure of the `coded_tests` spec group.

PR #202, opened 2024-09-19
Updated the pythonError codes to count down from 255.

Den1552:
## Summary

Related to #201 : Updated the pythonError codes to count down from 255.


aytey (approved):

Zbigor (approved):
all good

Issue #210, opened 2024-10-01
Coverage display in status bar should be improved


johnpaliotta:
Two issues

1. If there is one .c/.cpp editor file open, the status bar will display the coverage for that file.  If you subsequently close that file so that no editors are open the coverage in the status bar remains with the data for the last file closed.
2. If you open a random .c/.cpp file, the status bar will indicate "No Coverage Data" even if that file is not part of a VectorCAST environment.  We should show the status bar item only for files that are under test in an environment

PR #213, opened 2024-10-08
Added E2E tests for status bar improvements

Den1552:
## Summary

Tests fixed issue #210. Based on PR #211.

## Added
- Added a test to ensure that the status bar correctly updates its display when a file is closed. 
- Added a test to confirm that the status bar item is shown only for files currently under test in an environment.

### DIFF
--- a/tests/internal/e2e/test/specs/vcast.create_script_2_and_run.test.ts
+++ b/tests/internal/e2e/test/specs/vcast.create_script_2_and_run.test.ts
@@ -191,7 +191,7 @@ describe("vTypeCheck VS Code Extension", () => {
 
     console.log("Verifying absence of coverage status for quotes_example.");
 
-    // Verifying that there is no coverage status for quotes_example
+    // Verifying that there is no coverage status for quotes_example as it not instrumented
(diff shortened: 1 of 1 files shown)


tests/internal/e2e/test/specs/vcast.create_script_2_and_run.test.ts
```diff
@@ -170,6 +171,70 @@ describe("vTypeCheck VS Code Extension", () => {
     });
   });
 
+  it("should verify no coverage in status bar when opening quotes_example.cpp. ", async () => {
```
Zbigor, line 174: clearer: when opening uninstrumented quotes_example.cpp

Zbigor (approved):
Looks good

tests/internal/e2e/test/specs/vcast.create_script_2_and_run.test.ts
Den1552: Changed in a331feec08a94d4a0f35d05264aeacbad2967b4f .

PR #217, opened 2024-10-17
Clicast server release

johnpaliotta:
This PR has all of the changes that make the clicast server mode "releasable".

The benefit of server mode is much faster interaction with vcast environments.  For example, test execution takes a fraction of a second vs 2-3 seconds in non-server mode.  The time improvement is because server mode gets rid of the startup cost of vpython and clicast (initialization, license checkout, environment open, etc.)

Complete usage notes have been added to the README.md, but to summarize ...

- A new option has been added to "Use Data Server" --> default ON
- On startup, we check if the VectorCAST version that is configured for the workspace supports server mode (>vc24sp5)
- If the option is on and vcast supports server mode we start a server instance, and ping it to verify successful startup
- if startup is successful, we configure the extension to use server mode, and add a "vDataServer" button to status bar
- The button shows the current server state, and can be used to enable and disable the server.


src/vcastAdapter.ts
```diff
@@ -418,9 +355,11 @@ export async function runBasisPathCommands(
   testScriptPath: string,
   loadScriptCallBack: any
 ) {
-  // Execute Clicas tWith Progress() uses spawn() which needs the args as a list
+  // Execute Clicast With Progress() uses spawn() which needs the args as a list
   let argList: string[] = [];
   argList.push(`${clicastCommandToUse}`);
+  // note that we do NOT set the usingServer paraemter here
```
aytey, line 361: Typos

src/vcastAdapter.ts
```diff
@@ -464,6 +403,8 @@ export async function runATGCommands(
   // Execute Clicast With Progress() uses spawn() which needs the args as a list
   let argList: string[] = [];
   argList.push(`${atgCommandToUse}`);
+  // note that we do NOT set the usingServer paraemter here
```
aytey, line 406: Typos

aytey (approved):
LGTM

PR #218, opened 2024-10-18
Implemented HTTP POST in server mode.

Den1552:
## Summary

Previously, the client requests were sent using the fetch method with a GET request. This has now been updated to use axios.post.


src-common/vcastServer.ts
```diff
@@ -196,84 +196,55 @@ export async function transmitCommand(
...
-        // for some reason, when the server is down, the reason is blank
-        // so we insert a generic message in this case.
-        if (errorText.length === 0) {
-          errorText = `cannot communicate with server on port: ${SERVER_PORT}`;
-        }
-        let errorDetails = `command: ${requestObject.command}, error: ${errorText}`;
-        transmitResponse.success = false;
-        transmitResponse.statusText = `Enviro server error: ${errorDetails}`;
 
-        // In the core extension, this callback will shutdown the
```
aytey, line 271: Can we check what's happened with these comments, and if we've lost "value" here by losing these comments?

src-common/vcastServer.ts
```diff
@@ -196,84 +196,83 @@ export async function transmitCommand(
...
-        transmitResponse.success = true;
-        transmitResponse.statusText = `Enviro server response status: ${response.statusText}`;
-        // the format of the data property is different based on the command
-        // so it is up to the caller to interpret it properly
-        transmitResponse.returnData = rawReturnData;
-        // there is always an exit code field but it is only used when
-        // executing tests or running clicast commands
-      }
-    })
+    // Handle different exit codes from the server
```
aytey, line 208: response, not exit

src-common/vcastServer.ts
```diff
@@ -196,84 +196,83 @@ export async function transmitCommand(
...
+    // Await axios post
+    const response = await axios.post(urlToUse, dataAsString, {
+      headers: {
+        "Content-Type": "text/plain",
+      },
+    });
+    const rawReturnData = response.data;
 
-      if (rawReturnData.exitCode == pythonErrorCodes.internalServerError) {
-        // the error message is a list of strings, so join with \n
```
aytey, line 205: Missing this comment

python/vcastDataServer.py
```diff
@@ -25,27 +25,27 @@
...
+        @app.route("/runcommand", methods=["POST"])
         def runcommandRoute():
-            clientRequestText = request.args.get("request")
+            # Data from the request is a stringyfied json
+            clientRequestText = request.data.decode("utf-8")
             clientRequest = decodeRequest(clientRequestText)
+            # Ensure clientRequest is correctly decoded or processed
             return runcommand(clientRequest, clientRequestText)
 
-        # Note: this string must match what is in vcastAdapter.ts -> startServer()
```
johnpaliotta, line 45: Restore comment

python/vcastDataServer.py
```diff
@@ -25,27 +25,28 @@
...
+        @app.route("/shutdown", methods=["POST"])
         def shutdownRoute():
             return shutdown()
 
-        @app.route("/runcommand")
+        @app.route("/runcommand", methods=["POST"])
         def runcommandRoute():
-            clientRequestText = request.args.get("request")
+            clientRequestText = request.data.decode(
+                "utf-8"
```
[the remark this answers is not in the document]
Den1552: I wanted to keep the rest of the logic (e.g., decodeRequest(), etc.) unchanged, so I’m still sending the request as a string, like before. However, I can’t use `request.args.get("request")` with Axios because Axios sends its data in the request body, whereas the previous approach retrieved data from the query string `(http:...../?request`). I get this error 


`Enviro server error: client request was improperly formatted, ignoring: 'None'`

When doing: 

`clientRequestText = request.args.get("request")`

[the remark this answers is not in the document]
aytey: What's the type of `request.data`? I'm not questioning the removal of the `get`, I'm questioning why you need the `decode`.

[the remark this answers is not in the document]
aytey: Can we use `data = request.get_json()`, and then `decodeRequest` doesn't need to do the `json.loads`?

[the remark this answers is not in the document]
Den1552: Sure, we can do that. I just didn’t want to modify code where I wasn’t specifically asked to, or change the request format to JSON before it's wrong that I changed it.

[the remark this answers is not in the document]
Den1552: Changed request format in c9c845a8b485173cc75c581d64e8f10b58eeb68e.

src-common/vcastServer.ts
Den1552: Added comments back again.

src-common/vcastServer.ts
Den1552: Restored comment in c9c845a8b485173cc75c581d64e8f10b58eeb68e.

python/vcastDataServer.py
Den1552: Restored comment in c9c845a8b485173cc75c581d64e8f10b58eeb68e.

src-common/vcastServer.ts
Den1552: Changed in 008dd244ab1b3403497fe4f256c23ff43026b638.

PR #220, opened 2024-10-21
Got rid of the generation of the text version of the execution report

johnpaliotta:
Simplifies the code, and prepares for a new execution report generation method based on dataAPI

In the early days we were parsing the summary line in the text report "Expected Results matched 100% ( 1 / 1 )   PASS", but 
- We no longer do this 
- We would use the dataAPI if this is needed in the future.

### DIFF
--- a/python/clicastInterface.py
+++ b/python/clicastInterface.py
@@ -354,17 +354,11 @@ def generateExecutionReport(enviroPath, testIDObject):
     # We build a clicast command script to generate the execution report
     # since we need multiple commands
     with open(commandFileName, "w") as commandFile:
-        commandFile.write(
-            standardArgs
-            + " report custom actual "
-            + testIDObject.reportName
-            + ".html\n"
-        )
-        commandFile.write("option VCAST_CUSTOM_REPORT_FORMAT TEXT\n")
-        commandFile.write(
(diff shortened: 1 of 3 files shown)


python/clicastInterface.py
```diff
@@ -354,17 +354,11 @@ def generateExecutionReport(enviroPath, testIDObject):
     # We build a clicast command script to generate the execution report
     # since we need multiple commands
     with open(commandFileName, "w") as commandFile:
+        # we force report mode to HTML just to be safe
+        commandFile.write("option VCAST_CUSTOM_REPORT_FORMAT HTML\n")
```
aytey, line 358: I'm not saying this is better/worse than what we have currently, but won't this leave the environment in the HTML state if the user has selected TEXT?

johnpaliotta: Yes, as it did before.  I think that the number of people who have environments in text mode would be really small,
but we could open the CFG file to check the current value (if any) to improve this code.

I'm happy to do this, but we would have to do this file open/parse on each execute with report, or report command to be perfect.  Or we could do it once for each "base" directory when we load the environments.

When I initially wrote this code, my opinion was that it was not worth the effort.

aytey: As soon as we implement the new reporting approach, this will all go away anyway, so it really isn't a big deal to me ...

PR #248, opened 2024-12-13
Fixed wrong Test Results message log.

Den1552:
The logs displayed in the Test Results pane were previously unstructured and inconsistent compared to the formatting in the normal Output message pane. This PR addresses the issue by ensuring that the logs in the Test Results pane are properly formatted.

### DIFF
--- a/src/utilities.ts
+++ b/src/utilities.ts
@@ -305,3 +305,31 @@ export function removeFilePattern(enviroPath: string, pattern: string) {
     fs.unlinkSync(filePath);
   }
 }
+
+/**
+ * Cleans the message we want to show in the output. The Test Results pane handles logs differently.
+ * @param testResultString Test result we want to clean
+ * @returns Cleaned message, ready for the Test Results pane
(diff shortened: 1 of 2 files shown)


src/utilities.ts
```diff
@@ -305,3 +305,31 @@ export function removeFilePattern(enviroPath: string, pattern: string) {
...
+  let alignedOutput = cleanedOutput
+    .map((line, index) => {
+      // First line stays unmodified
+      if (index === 0) {
+        return line.trim();
+      }
+      // Apply second line padding to subsequent lines
+      return secondLinePadding + line.trim();
+    })
+    .join("\r\n");
```
aytey, line 332: Why `\r`?

Den1552: If you hover over the run.appentOutput you will see this info message: 

`[...] New lines must be given as CRLF (\r\n) rather than LF (\n).`

Zbigor (approved):

PR #249, opened 2024-12-18
Updated Tst autocompletion support for ct (Issue #170)

### LINKED ISSUE #170: Test script editing needs some updates to support coded tests

Den1552:
## Summary

Updates script editing to support coded tests and resolves #170. 

## Added
- **Syntax Support for `TEST.CODED_TEST_FILE`:**
  - Added `TEST.CODED_TEST_FILE` as a valid syntax in `.tst` files.
  - Enabled auto-complete for `TEST.CODED_TEST_FILE` after typing `TEST.`.
  - Ensured no auto-completion for the file path after `:`.

- **Behavior When `TEST.SUBPROGRAM` is `coded_tests_driver`:**
  - Allowed `TEST.CODED_TEST_FILE` without error diagnostics.
  - Disallowed `TEST.VALUE` and `TEST.EXPECTED` lines, ensuring they give error diagnostics.

- **Behavior When `TEST.SUBPROGRAM` is Not `coded_tests_driver`:**
  - Disallowed `TEST.CODED_TEST_FILE`, ensuring it gives an error diagnostic.
  - Removed `coded_tests_driver` from the subprogram list for `TEST.VALUE` and `TEST.EXPECTED` lines.

- **Error Handling Enhancements:**
  - Reviewed and ensured consistency in error messages related to these changes.

- **Tests**
   - Added tests for new diagnostics

- **Autocompletion list for `TEST.SUBPROGRAM`** without using the fake `TEST.VALUE` line 
  - In order to get the autocompletion list for `TEST.SUBPROGRAM`, we need the `<unit>` from `TEST.UNIT`.
  - This is retrieved with `getNearest()` in `tstCompletion.ts`.
  - This `<unit>` is sent as an additional parameter to the Python logic.
  - In `processSubprogramLines()` (tstUtilities.py), we use this `<unit>` to retrieve, with `getFunctions()`, the `autocompletion` list of all the functions in this unit from the `DataAPI`.
  - At the end, we add SUBPROGRAM-specific autocompletions like `<<INIT>>, <<COMPOUND>>, <<coded_tests_driver>>`.

### DIFF
--- a/langServer/tstCompletion.ts
+++ b/langServer/tstCompletion.ts
@@ -17,9 +17,64 @@ import {
   emptyChoiceData,
   getChoiceData,
 } from "./pythonUtilities";
+import { promisify } from "util";
+import { exec } from "child_process";
 
 function filterArray(currentArray: string[], whatToRemove: string) {
   return currentArray.filter((element) => element !== whatToRemove);
+}
+
+/**
+ * Checks if a specific VectorCAST option is set to a given value in the specified environment.
+ *
+ * @param {string} enviroPath - The file path to the environment where the command should be executed.
+ * @param {string} option - The VectorCAST option to check (e.g., 'VCAST_CODED_TESTS_SUPPORT').
+ * @param {string} optionValue - The expected value of the option (e.g., 'True', ...).
+ * @returns {Promise<boolean>} - Resolves to `true` if the option value matches, otherwise `false`.
+ */
+export async function checkClicastOption(
+  enviroPath: string,
+  option: string,
+  optionValue: string
+): Promise<boolean> {
+  const execAsync = promisify(exec);
+  const getCodedTestsSupportCommand = `${process.env.VECTORCAST_DIR}/clicast option ${option}`;
+
+  try {
+    const { stdout } = await execAsync(getCodedTestsSupportCommand, {
+      cwd: enviroPath,
+    });
+
+    return stdout.includes(`${optionValue}`);
+  } catch (stderr) {
+    console.error(`Error executing command: ${stderr}`);
+    return false;
+  }
+}
+
+/**
+ * Checks if a specific keyword followed by a colon and a given value exists in the extracted text.
+ *
+ * @param {string} extractedText - The text to search within.
+ * @param {string} keyword - The keyword to search for.
+ * @param {string} value - The value that should follow the colon after the keyword.
+ * @returns {Promise<boolean>} - Resolves to `true` if the keyword and value are found in the text, otherwise `false`.
+ */
+export function checkForKeywordInLine(
+  extractedText: string,
+  keyword: string,
+  value: string
+): boolean {
+  const regex = new RegExp(`^${keyword}:${value}$`);
+  const lines = extractedText.split(/\r?\n/);
+
+  for (const line of lines) {
+    if (regex.test(line.trim())) {
+      return true;
+    }
+  }
+
+  return false;
 }
 
 export async function getTstCompletionData(
@@ -29,6 +84,17 @@ export async function getTstCompletionData(
   let returnData: choiceDataType = emptyChoiceData;
   const testScriptPath = url.fileURLToPath(completionData.textDocument.uri);
   const enviroPath = getEnviroNameFromTestScript(testScriptPath);
+  const extractedText = currentDocument.getText();
+
+  let codedTestsEnabled;
+
+  if (enviroPath) {
+    codedTestsEnabled = checkClicastOption(
+      enviroPath,
+      "VCAST_CODED_TESTS_SUPPORT",
+      "True"
+    );
+  }
 
   if (enviroPath && fs.existsSync(enviroPath)) {
     // The work we do is dependent on the trigger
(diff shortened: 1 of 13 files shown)


langServer/tstCompletion.ts
```diff
@@ -70,21 +174,23 @@ export async function getTstCompletionData(
...
     } else if (trigger == "COLON" && upperCaseLine == "TEST.SUBPROGRAM:") {
       // find closest TEST.UNIT above this line ...
-      const unitName = getNearest(
+      const unit = getNearest(
         currentDocument,
         "UNIT",
         completionData.position.line
       );
+
       // TBD will need to change how this is done during the fix for issue #170
```
Zbigor, line 183: we can remove this comment now :)

langServer/tstValidation.ts
```diff
@@ -48,12 +50,17 @@ export function validateTextDocument(textDocument: TextDocument) {
...
   let withinImportFailures = false;
+  const codedTestsDriverInSubprogram = checkForKeywordInLine(
+    text,
+    "TEST.SUBPROGRAM",
+    "coded_tests_driver"
+  );
 
-  for (let lineIndex = 0; lineIndex < lineList.length; lineIndex++) {
+  for (lineIndex = 0; lineIndex < lineList.length; lineIndex++) {
     let thisLine: string = lineList[lineIndex];
```
Zbigor, line 60: Not suggesting to do this rignt now, but line 61 - line 170 could be refactored into clearer statements with less nesting, such that it's clearer which particular case is being handled. Could be turned into an "internal issue" for later

Zbigor: It would make sense to cover the currently uncovered lines in this range with unit tests

python/tstUtilities.py
```diff
@@ -1039,6 +1059,12 @@ def processTstLine(enviroPath, line):
             returnData = processSlotLines(api, pieces, triggerCharacter)
         elif line.upper().startswith("TEST.REQUIREMENT_KEY"):
             returnData = processRequirementLines(api, pieces, triggerCharacter)
+        elif line.upper().startswith("TEST.SUBPROGRAM"):
+            if unit == None:
+                globalOutputLog.append(
+                    "Additional 'unit' parameter is requiered for TEST.SUBPROGRAM: autocompletion."
```
Zbigor, line 1065: required

Zbigor (changes requested):
Looks alright. It would make sense to add a couple more e2e tests for the completion itself (suggested unit/function) and to cover that wider range indicated by codecov (the single line special failure cases aren't as important). The comment about a possible refactor could be turned into an "internal issue"

python/tstUtilities.py
Den1552: Changed in 265346f47f3d1a2bda59ea03cf5c5e455c889ea0

langServer/tstValidation.ts
Den1552: Brought unit tests back to 100% in 265346f47f3d1a2bda59ea03cf5c5e455c889ea0

langServer/tstCompletion.ts
Den1552: Removed in 265346f47f3d1a2bda59ea03cf5c5e455c889ea0.

Zbigor (approved):

PR #270, opened 2025-07-31
Env loading performance

Den1552:
Enhances initial env loading performance by loading all env info at once in one vPython process

### DIFF
--- a/python/vTestInterface.py
+++ b/python/vTestInterface.py
@@ -769,8 +769,10 @@ def processCommandLogic(mode, clicast, pathToUse, testString="", options=""):
 
     elif mode == "getWorkspaceEnviroData":
         enviro_list = []
+        errors = []
         topLevel = {}
         vce_files = find_vce_files(pathToUse)
+
         for vce_path in vce_files:
             try:
                 api = UnitTestApi(vce_path)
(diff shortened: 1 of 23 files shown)


src/testPane.ts
```diff
@@ -829,34 +970,38 @@ async function loadAllVCTests(
...
-
+    if (environmentList.length > 0) {
+      vcastEnvironmentsFound = true;
+    }
+    const increment = 100 / environmentList.length;
     for (const environmentData of environmentList) {
+      if (token.isCancellationRequested) {
+        cancelled = true;
+      }
+      if (cancelled) {
```
aytey, line 981: Why not just `if (token.isCancellationRequested)`?

python/vTestInterface.py
```diff
@@ -752,6 +768,40 @@ def processCommandLogic(mode, clicast, pathToUse, testString="", options=""):
...
+                        "unitData": unit_data,
+                        "mockingSupport": mocking_support,
+                    }
+                )
+
+            except Exception as err:
+                errors.append(f"{vce_path}: {str(err)}")
+
+        topLevel["testData"] = enviro_list[0]["testData"] if enviro_list else []
+        topLevel["unitData"] = enviro_list[0]["unitData"] if enviro_list else []
```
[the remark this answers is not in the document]
Den1552: I know the structutre because I literally build it a few lines above? 

```
enviro_list.append(
    {
        "vcePath": vce_path,
        "testData": test_data,
        "unitData": unit_data,
        "mockingSupport": mocking_support,
    }
)
```

src/testPane.ts
Den1552: Probably because initially I had more information messages in that function where I could cancel in each one of them. Changed.

PR #283, opened 2025-08-25
Tst deletion fix

Den1552:
This PR fixes an issue where manually edited .tst scripts were being deleted after saving. Previously, all loaded scripts were unlinked regardless of their origin.


Zbigor (approved):
LGTM

PR #289, opened 2025-09-19
Multiple workspacefolders fix

Den1552:
- Previously, when opening multiple workspace folders in a single VS Code session, the extension only scanned .vce files from the first folder and ignored all others. 

- Removed some redundant backend return.


Zbigor (approved):
lgtm

PR #295, opened 2025-10-08
Small fixes regarding high-level requirement generation

ThisCakeIsALie:
Corrects the config location used for determining if we should generate high-level requirements and adds the requirement module/unit as a backup group to group requirements under before falling back to "Unknown Function".


Zbigor (approved):
Isolated fix, can go in immediately

PR #298, opened 2025-10-27
Adapted CI workflow

Den1552:
Removing reqs2test creation scripts and refactored workflow for the vsix download.


Zbigor (approved):
Let's change the name to "create-extension" or "build-extension" as this pipeline doesn't actually publish to the marketplace, just to be explicit. Cause if we wanted, we could actually automate that as well

PR #301, opened 2025-11-10
Fixed reqs issue with env var in path && reorganized reqs code

Den1552:
Expands env variables from the CCAST.cfg when doing reqs operations and reorganized requirements code for clarity

### DIFF
--- a/src/requirements/requirementsOperations.ts
+++ b/src/requirements/requirementsOperations.ts
@@ -177,7 +177,7 @@ export async function generateRequirements(enviroPath: string) {
     "--export-excel",
     xlsxPath,
     "--export-repository",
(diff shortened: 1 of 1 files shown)


src/requirements/requirementsUtils.ts
```diff
@@ -0,0 +1,568 @@
...
+  return spawn(command, args, { ...options, env });
+}
+
+export function expandEnvVars(inputPath: string): string {
+  return inputPath.replace(/\$\(([^)]+)\)/g, (match, varName) => {
+    const value = process.env[varName];
+    // Leave it unchanged if not found
+    return value ? value : match;
+  });
+}
```
Den1552: Helper function to expand env vars

PR #307, opened 2025-12-02
Vcshell implementation

Den1552:
Adds: 

- Ability to set a VCDB as default and include in the default CFG
- Ability to set VCDB option when creating new Projects / CFG files
- Redesigned new Project / new CFG webviews


src/manage/webviews/html/newProject.html
```diff
@@ -25,10 +25,86 @@ <h2>Create New Project</h2>
...
+      <h3>Compiler Selection</h3>
+
+      <!-- Only shown if defaultCFG is available -->
+      <div id="defaultCompilerRow" class="selection-box" style="display:none;">
+        <label class="toggle-label" for="useDefaultCompiler">
+          Use Default CFG Compiler<br />
+          <small id="defaultCFGPath"></small>
+        </label>
+        <label class="vscode-toggle" for="useDefaultCompiler">
+          <span class="toggle-switch">
```
Zbigor: worth addresing

Zbigor (changes requested):
No major issues, but it'd make sense to look at the following:
- Sonar findings, not many and they seem to make sense
- Potential code duplication (we might have a reason why we want to duplicate?)
- Is it possible to add any unit tests for the new logic?
  - And/Or some e2e at least

PR #312, opened 2025-12-10
C Coverage 

Den1552:
Adds Coverage for Cover Project Files in VSCode. 

Main Changes: 
- Finding vcp files
- Showing Cover projects as nodes that are not interactable
- Files of Cover Projects now have Coverage too
- Mcdc Report Adaptions for Cover Projects


Zbigor (changes requested):
looks alright, just the 2 points:
- Can we add an e2e test at least
- Minor refactoring from comments

PR #317, opened 2026-01-23
Sourcefile open from testing pane

Den1552:
Adds the Feature to open up a Source File from the Testing pane by either clicking on the unit or clikcing on a function within the units that also opens up the source file and focuses on where the function starts

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -68,7 +68,6 @@ import {
   setGlobalProjectIsOpenedChecker,
   setGlobalCompilerAndTestsuites,
   loadTestScriptButton,
-  cachedWorkspaceEnvData,
 } from "./testPane";
 
 import {
@@ -90,7 +89,6 @@ import {
   rebuildEnvironment,
   openProjectInVcast,
   deleteLevel,
-  getDataForEnvironment,
 } from "./vcastAdapter";
 
 import {
@@ -157,6 +155,7 @@ import {
   addIncludePath,
   envIsEmbeddedInProject,
   getEnviroNameFromFile,
+  getEnvironmentData,
   getLevelFromNodeId,
   getVcmRoot,
   openProjectFromEnviroPath,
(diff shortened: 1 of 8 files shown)


src/extension.ts
```diff
@@ -1235,6 +1237,98 @@ function configureExtension(context: vscode.ExtensionContext) {
...
+                cacheKey as keyof typeof cachedWorkspaceEnvData
+              ];
+          }
+
+          // If not in cache, fetch it
+          if (!envData) {
+            envData = await getDataForEnvironment(enviroPath);
+          }
+
+          if (envData && envData.unitData) {
```
Zbigor: Could you try reducing the nesting here a bit? For example by returning on opposite conditions etc. It's still understandable what's going on, so this is not so critical

tests/internal/e2e/test/test_utils/vcast_utils.ts
```diff
@@ -1581,30 +1581,49 @@ export async function checkForGutterAndGenerateReport(
...
+      break;
+    }
   }
-  // Check if the file is already open in the editor
+
+  if (!unit) {
+    throw new Error(`Unit ${unitName} not found in Testing pane`);
+  }
+
+  // Open the source file through the context menu
```
Zbigor: I think it's fine that it's tested like this for now, cause there are multiple instances of stuff being tested like this. But there could be a refactor that'd make it clearer which test is verifying which feature/user flow - a separate one that would address all such instances. Here it'd be clear from the screenshot of the test failure which flow actually failed, so it's fine for now

Zbigor (approved):

tests/internal/e2e/test/test_utils/vcast_utils.ts
Den1552: Refactored and isolated as an own test

src/extension.ts
Den1552: Refactored a little bit and made a util to make it smaller

PR #320, opened 2026-02-24
Improved error handling during initial buildup of workspace cache

Zbigor:
The VS-Code extension was crashing on startup for a customer. The error could be traced to accessing `.error` on an undefined object that was created as a response to a workspace environment query. 

We are now logging the exact workspace directory (one of potentially many) where the `vpython` command results in an undefined response. We now filter out undefined responses and make sure to skip them before trying to access the `.error` attribute.

### DIFF
--- a/src/utilities.ts
+++ b/src/utilities.ts
@@ -430,6 +430,7 @@ export async function mergeWorkspaceEnvResponses(
   const allEnvs: EnviroData[] = [];
 
   for (const resp of responses) {
+    if (!resp) continue;
     if (resp.errors) {
       allErrors.push(...resp.errors);
     }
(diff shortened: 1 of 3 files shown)


src/utilities.ts
```diff
@@ -430,6 +430,7 @@ export async function mergeWorkspaceEnvResponses(
   const allEnvs: EnviroData[] = [];
 
   for (const resp of responses) {
+    if (!resp) continue;
```
Den1552, line 433: Isn't this dead code if we filter the undefined / null responses in 

https://github.com/vectorgrp/vector-vscode-vcast/pull/320/changes#diff-97135952eb8a8df8a83efdd57970ae0a36dc0456715b3d74a08afc4a45d5a9c2R782-R785

?

Den1552 (commented):
LGTM, except that I think we can remove the line I commented.

src/utilities.ts
Zbigor: I really wanted to be sure and explicit here

PR #329, opened 2026-04-28
Switch requirements storage backend in the extension to RGW

ThisCakeIsALie:
Switches the backend to read from and work with the RGW for the VectorCAST environment directly. Also adds a webview that allows modification of requirements as well as other tweaks (greying out buttons instead of hiding them, making requirements traceability more intuitive, etc.)


src/requirements/webviews/webviewScripts/requirements.js
```diff
@@ -0,0 +1,116 @@
...
+saveBtn.addEventListener("click", postSave);
+inferBtn.addEventListener("click", postInfer);
+
+if (searchInput) searchInput.addEventListener("input", applyFilter);
+if (filterUnit) filterUnit.addEventListener("change", applyFilter);
+if (filterFunction) filterFunction.addEventListener("change", applyFilter);
+
+// ---------- Extension → webview messages ---------------------------------
+
+window.addEventListener("message", (event) => {
```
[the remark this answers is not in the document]
ThisCakeIsALie: Fixed

docs/reqs2x/reqs2x_documentation.md
```diff
@@ -94,21 +93,18 @@ Remove any existing versions of the VectorCAST VS-Code extension.
...
 
 
 ### Initial Setup
 1.  Click the **Flask icon** (Test Explorer) on the left sidebar to show the environment tree.
-2.  Right-click `TUTORIAL_C` and select **VectorCAST -> Show Requirements**. The requirements webview will appear.
-3.  Right-click `TUTORIAL_C` and select **VectorCAST -> Populate RGW from Requirements**.
-    *   *Note: This creates a requirements gateway and populates it from the Excel sheet. Wait for the notification (approx. 3-5 seconds).*
-    *   If you do not have the requirements-to-code traceability for a function, Reqs2x will generate it automatically with your approval.
+2. Right-click `TUTORIAL_C` and select **VectorCAST -> Import Requirements**. Choose the file `reqs-TUTORIAL_C/reqs.xlsx`. This will load the requirements into the environment.
+2.  Right-click `TUTORIAL_C` and select **VectorCAST -> Show Requirements**. The requirements webview will appear, rendered directly from the RGW.
```
Den1552: 2 x number 2

src/requirements/webviews/webviewScripts/filter.js
```diff
@@ -0,0 +1,163 @@
...
+export function rebuildFilterDropdowns() {
+  if (!filterUnit || !filterFunction) return;
+
+  const units = new Set();
+  const fns = new Set();
+
+  const map = state.unitsToFunctions || {};
+  for (const u of Object.keys(map)) {
+    units.add(u);
+    for (const f of map[u] || []) fns.add(f);
```
Den1552: Can we maybe transform / add these var names in the files? So here for instance f -> function. u -> unit. Maybe a little bit picky but I think claude can handle it quickly and it's easier to read imo.

src/requirements/webviews/messages.ts
Den1552, line 1: 2 Things I would like to suggest here: 

1) Can we maybe move those Scripts into the webviews folder? I just think it's a little bit odd to have a webview and a webviews folder in the same folder. 
2) Can we extract the html from template.ts and put it in a webvies/html folder? Similar like we do in manage/webviews. We probably will need to add more webviews in the future around the requirements feature and just to have it consistent with the other folder strucutre + maintainability.

src/extension.ts
```diff
@@ -1244,126 +1394,204 @@ function configureExtension(context: vscode.ExtensionContext) {
...
 
   let showRequirementsCommand = vscode.commands.registerCommand(
     "vectorcastTestExplorer.showRequirements",
     async (args: any) => {
-      if (args) {
-        const testNode: testNodeType = getTestNode(args.id);
-        const enviroPath = testNode.enviroPath;
+      if (!args) return;
+      const testNode: testNodeType = getTestNode(args.id);
+      const enviroPath = testNode.enviroPath;
```
Den1552: Null Check like in resolveEnviroPathForCommand?

src/extension.ts
```diff
@@ -1385,63 +1613,46 @@ function configureExtension(context: vscode.ExtensionContext) {
...
-                );
-              }
+          const gatewayPath = findRelevantRequirementGateway(enviroPath);
+          if (gatewayPath && fs.existsSync(gatewayPath)) {
+            try {
+              fs.rmSync(gatewayPath, { recursive: true, force: true });
+            } catch (err) {
+              vscode.window.showErrorMessage(
+                `Failed to remove requirements gateway: ${err}`
+              );
```
Den1552: I think this causes a bug. When I wrote the tests and tried to remove the imported RGW form the test setup, far more got deleted than intended and the environment no longer showed up afterwards (see screenshot).
I think doing an unguarded` fs.rmSync(gatewayPath, { recursive: true, force: true })` on whatever **VCAST_REPOSITORY** points at, with no check that the path is actually contained inside the env directory causes the issue. In this case the RGW wasn't created where we expect (reqs-#env#/rgw)

Can we maybe resolve and expand the path, refuse to delete unless it's strictly inside the env's parent dir ?

Image

Den1552 (changes requested):
Added a few comments. YOu might also take a look at the sonar findings + run npx prettier -w <ts & js -files>. Prettier should be installed if you did npm install on the extension. Also please update in package.json the version number and add a Changelog for the new release. Otherwise LGTM

ThisCakeIsALie:
Issues should all be addressed

PR #337, opened 2026-06-29
Rebuild Fail Delete fix

Den1552:
Falls back to the original env state when an env rebuild fails. Previously, a failed rebuild removed parts of the build directory.


Zbigor (approved):
lgtm

PR #344, opened 2026-07-29
Expose recent Reqs2X flags, add delta test generation, wire up the nele provider

ThisCakeIsALie:
Catches the extension up with flags and events the Reqs2X CLI has gained.

**Settings** — `batchSize`, `fastMode`, `deduplicateTests`, `allowPartialTests`, `maxTestExamples`, `testExampleSources`. `--allow-partial` was hardcoded on with no way to disable it. `noTestExamples` is deprecated in favour of `maxTestExamples: 0` rather than removed, so existing configs keep working (note VS Code hides deprecated settings unless they have a value).

**Delta generation** — second command, *Generate Tests from Requirements (Untested or Changed Only)*, for `--only-delta`. A command rather than a setting because it is per-run intent. Its menu entry is gated on a context key so it stays hidden when the installed Reqs2X predates the flag. A delta run with nothing to do writes no `.tst`, so the target is removed up front and its absence tolerated afterwards — otherwise the previous run's script gets re-imported.

**nele provider** — was supported by the CLI but not selectable.

**Event handling** — `ProgressTracker` dropped `info` and `notice`. `info` already fires on the `--only-untraced` path we use, so inferring traceability on a fully-traced RGW finished with no visible output. `notice` carries the subscription nudge.

**`--no-reorder` removed** — no released CLI has ever accepted it, so turning the `reorder` setting off made code2reqs and reqs2tests exit non-zero.

**Version skew** — `panreqSupportsOnlyUntraced` was a bespoke single-flag cache; now `reqs2xSupportsFlag` / `reqs2xSupportedFlags` over one cached `--help` per binary. Recent flags are probed and dropped when absent; because settings cannot be gated on a context key the way a command can, generation warns when a setting the user actually chose was dropped.

Also drops the per-event log line from the availability refresh, which the `CCAST_.CFG` watcher was emitting several times a second while clicast was active.

### DIFF
--- a/package.json
+++ b/package.json
@@ -509,12 +509,6 @@
             "default": true,
             "description": "Generate tests in batches (multiple requirements per LLM call). Disable for one-request-per-requirement (slower, simpler retries)."
           },
-          "vectorcastTestExplorer.reqs2x.reorder": {
-            "type": "boolean",
-            "order": 8,
-            "default": true,
-            "description": "Reorder test cases exported to file (disable for easier investigation of incremental results)"
-          },
           "vectorcastTestExplorer.reqs2x.noTestExamples": {
             "type": "boolean",
             "order": 9,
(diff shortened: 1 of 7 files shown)


Den1552 (approved):
I only had format the ts files with prettier but otherwise LGTM. Tests also pass.

PR #345, opened 2026-08-19
CSV RGW gateways now work correctly when inferring traceability and modifying them on VC26-SP1 and beyond

ThisCakeIsALie:
The layout of the CSV gateway settings changed. This PR makes looking up the relevant settings robust across both layouts.


Den1552:
I removed some unused imports + prettier, otherwise LGTM.

